PR #25, opened 2022-04-22
feat: implement `GITHUB_PR_COMMITS` sync type

alonlong:


internal/syncer/github_pr_commits.go
```diff
@@ -0,0 +1,153 @@
...
+			return fmt.Errorf("uuid: %w", err)
+		}
+
+		key := fmt.Sprintf("%d-%s", *r.PRNumber, *r.Hash)
+		if duplicated[key] {
+			w.logger.Warn().Msgf("duplicated input: %s, %s", repoID.String(), key)
+			continue
+		} else {
+			duplicated[key] = true
+		}
```
patrickdevivo, line 69: let's remove this duplicated check for now for consistency with the other handlers. I will make a note, in the future I think we can try a technique like so: https://stackoverflow.com/questions/48019381/how-postgresql-copy-to-stdin-with-csv-do-on-conflic-do-update (copy into a temp table, then insert from there) to ignore any duplicates gracefully. This could be a better way in general to do the batch inserts

alonlong: done

patrickdevivo (approved):

PR #67, opened 2022-07-08
Auth token

emilyvandewalle:
Side note: I've added an `isEmpty` variable to test the initial state as I wasn't sure how to trigger it


patrickdevivo:
@emilyvandewalle I see the `Validate` button is below the input, could we move it to the right similar to match what's in Figma?

image

image

emilyvandewalle:
Correct, the disabled state should already be merged into blocks when adding the `VerticalNav` component ([link](https://624b441a141307004a094f09-pivkvhkqdz.chromatic.com/?path=/docs/molecules-verticalnav--vertical-nav-with-icons)), I'll merge in the mono font changes, update the package and update the settings nav here 👍

emilyvandewalle:
@patrickdevivo It should be updated now

PR #85, opened 2022-07-28
fix: add a longer timeout to the `golanci-lint` step

patrickdevivo:
in the GitHub action. we frequently trip the default timeout on mac os builds. Hopefully this helps avoid it.


amenowanna (approved):
Nice! I hit this issue the other day too.

PR #94, opened 2022-08-01
feat: start experiment to enable custom SQL in GraphQL

### LINKED ISSUE #87: Implement backend for direct SQL querying access (for use via UI)
Similar to what we've done in `app.mergestat.com`, we'd like to make it possible (in a `Queries` section of the UI) to execute SQL directly against the data synced by Fuse.

This issue is about implementing the "backend" for this functionality (in Postgres/Graphile?)

patrickdevivo:
Extends the Graphile schema adding a very simple `execSQL` JS resolver (as a plugin) which executes a user-supplied query and returns the results as JSON.

This is in an effort to support a query page in the management UI.

- [x] Use TypeScript for type safety?
- [x] Support query variables
- [x] Update `docker-compose.yaml` to include plugin
- [ ] ~Implement some sort of fixed timeout?~
- [x] Implement some sort of row limiting?
- [x] Figure out how to make query execution read-only

Also - consider whether this could live as a standalone graphile plugin in a separate repo?

See here for a reference: https://github.com/graphile/pg-simplify-inflector/blob/main/package.json

Closes #87


graphql/exec-sql/tsconfig.json
```diff
@@ -0,0 +1,103 @@
...
+    // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
+    // "tsBuildInfoFile": "./.tsbuildinfo",              /* Specify the path to .tsbuildinfo incremental compilation file. */
+    // "disableSourceOfProjectReferenceRedirect": true,  /* Disable preferring source files instead of declaration files when referencing composite projects. */
+    // "disableSolutionSearching": true,                 /* Opt a project out of multi-project reference checking when editing. */
+    // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
+
+    /* Language and Environment */
+    "target": "es2016",                                  /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */
+    // "lib": [],                                        /* Specify a set of bundled library declaration files that describe the target runtime environment. */
```
german-mergestat, line 15: just a simple comment... I think all this comments blocks are not necessary because we can see all possible flags here in the documentation. I mean, just leaving this comment is enough

`/* Visit https://aka.ms/tsconfig to read more about this file */`

patrickdevivo: makes sense! thanks for that - update on the way

german-mergestat (approved):

amenowanna (approved):

Issue #69, opened 2022-07-08
Connect the UI to the backend for the repo info page


patrickdevivo:
Connect the following page in the UI to the GraphQL API

<img width="2048" alt="image" src="https://user-images.githubusercontent.com/57259/178010115-9c28fd88-2bb1-404b-8e11-e36f78c19644.png">

See [this query](https://github.com/mergestat/fuse/issues/64#issuecomment-1167433780) as a starting point

PR #97, opened 2022-08-01
Polling data for near real time

german-mergestat:
Adding near real time (Polling each 5 sec) in: 

-  **Repo List**
-  **Repo Syncs list.** 

(Complementing #69)

### DIFF
--- a/ui/src/pages/repos/[repository]/index.tsx
+++ b/ui/src/pages/repos/[repository]/index.tsx
@@ -1,4 +1,4 @@
-import { Fragment, useState } from 'react'
+import { Fragment } from 'react'
 import Head from 'next/head'
 import RepoDataView from 'src/views/repository-data'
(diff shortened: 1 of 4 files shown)


ui/src/pages/repos/[repository]/index.tsx
```diff
@@ -14,6 +14,7 @@ const RepoDetailsPage = () => {
 
   const { loading, error, data } = useQuery(GET_REPO_SYNCS, {
     variables: { id: repository },
+    pollInterval: 5000,
   })
```
Comment (patrickdevivo) [inline ui/src/pages/repos/[repository]/index.tsx:18]: we should probably handle the `error` case here as well, yeah? It can be simple for now - maybe we could use the alert stack? https://624b441a141307004a094f09-anonmxrnpj.chromatic.com/?path=/story/utilities-alertstack--alert-stack-util

patrickdevivo (approved):

PR #130, opened 2022-08-15
feat: add `/api/graphql` to proxy to graphql API

patrickdevivo:
This adds a an API route at `/api/graphql` which proxies requests to `process.env.POSTGRAPHILE_API`. This allows us to configure the server side Node.js with a standard env variable (that can be modified at runtime) and the browser will always call `/api/graphql` (no need to pass a config value all the way thru to the browser)


german-mergestat (approved):
I like it :+1:

PR #135, opened 2022-08-17
feat(search): Implementing repo search filter

### LINKED ISSUE #98: Enable search bar functionality in `repos` page (client side only)
Implement a client-side only search filter on the `repos` page. Match against the repo name.

In the future we may move this filtering to the server, definitely when we implement pagination.

image

german-mergestat:
Implementing repo search filter. (Resolves #98)

- Adding `unit test` for searching `angular` repository.

filter

### DIFF
--- a/ui/src/views/repositories/index.tsx
+++ b/ui/src/views/repositories/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
 import { useRepositoriesContext, useRepositoriesSetState } from 'src/state/contexts/repositories.context'
 import { EmptyRepositoryTable, FilterHeader, PageHeader, RepositoriesTable } from './components'
 import { SyncDataDropDown } from './drop-downs/sync-repos-data-drop-down'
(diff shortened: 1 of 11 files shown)


ui/src/views/repositories/index.tsx
```diff
@@ -14,16 +14,27 @@ import Loading from 'src/components/Loading'
...
 
-  // - Todo: connect selectedRepositoriesCount from RepositoriesTable
-  const selectedRepositoriesCount = 0
+  useEffect(() => {
+    refetch({ search })
+  }, [refetch, search])
+
+  if (data && !showTable) {
+    setShowTable(true)
+  }
```
patrickdevivo, line 37: this seems a little _funky_ 😃 to me - calling a `setXxx` function in the render method - feels like an antipattern? But I'm not entirely sure. Should it be in a `useEffect` or should we derive whether to show the table in the returned  JSX?

german-mergestat: mmm yeah, it looks some weird but I had to left it like so. I had it inside a `useEffect` at first but its behavior was no so good because it rendered for a short time `<Empty Repository Table />`, that's because it had to wait for `useQuery` to be resolved. Not sure if it is an anti-pattern but at the end was the best way to get a good user experience. The good thing is that `setShowTable(true)` is executed just one time (at first render)

patrickdevivo (approved):

ui/src/views/repositories/index.tsx
german-mergestat: Done :smiley: [89a363c](https://github.com/mergestat/fuse/pull/135/commits/89a363c1e1213ece491ae06301b73efa408ba1fc)

PR #137, opened 2022-08-18
fix: bug with where rate limit could be depleted

### LINKED ISSUE #134: Fuse worker depleting GitHub api rate limit quota
While load testing I came across a bug where the logic that is in place to ensure we don't deplete the API quota fails.

It appears that when we reach our 400 quota buffer we back off and wait as expected

Image

However the wait is not long enough and when the wait is released the logic no longer waits and allows the quota to be depleted.

Image

amenowanna:
- changed logic to delay the until the quota is reset
- added support for shutdown during the delay.

resolves #134


patrickdevivo (approved):
woohoo! thanks for taking care of this 🎉

PR #146, opened 2022-08-19
fix(user-icon): Removing user icon menu for now

### LINKED ISSUE #106: Remove user/account management button for now in UI
image

Since we have no concept of auth/authz yet, remove this from the UI for now

german-mergestat:
Removing user icon menu. (Resolves #106)

user


patrickdevivo (approved):
lgtm! I think in the future it would be good to split this up into 2 PRs for each piece of fucntionality:

- remove account UI
- upgrade codegen

PR #164, opened 2022-08-30
feat: enforce queue id order

amenowanna:


patrickdevivo:
@amenowanna added a commit that runs `sqlc generate` which should bring your SQL changes into the go package that's used

amenowanna:
@patrickdevivo muchas gracias!

ramiroJCB (approved):

PR #168, opened 2022-08-30
Add repo by csv

### LINKED ISSUE #100: Implement add repo flow for adding a CSV list of repos
image

german-mergestat:
Implementing repo addition by CSV. (Resolves #100)

add-csv

added-csv

### DIFF
--- a/ui/src/views/repositories/modals/add-repository-modal/add-repository-from-csv-modal.tsx
+++ b/ui/src/views/repositories/modals/add-repository-modal/add-repository-from-csv-modal.tsx
@@ -1,26 +1,17 @@
 import { Label, Textarea } from '@mergestat/blocks'
-import React, { ChangeEvent, useCallback, useEffect } from 'react'
+import React, { ChangeEvent, useEffect } from 'react'
 import { useRepositoriesSetState } from 'src/state/contexts'
 import { TEST_IDS } from 'src/utils/constants'
 
 export const AddRepositoryFromCSVModal: React.FC = () => {
-  const { setReposToAdd } = useRepositoriesSetState()
+  const { resetValues, setCSVText } = useRepositoriesSetState()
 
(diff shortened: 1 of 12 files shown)


ui/src/views/repositories/modals/add-repository-modal/add-repository-from-csv-modal.tsx
```diff
@@ -1,16 +1,36 @@
...
+    const matrix = lines.map(line => line.split(',').map(ele => ele.trim()))
+
+    let newRepos: string[] = []
+    matrix.forEach(line => {
+      if (line[0] !== '' && !newRepos.includes(line[0])) {
+        newRepos = [...newRepos, line[0]]
+      }
+    })
+    setReposToAdd(newRepos)
+  }, [setReposToAdd])
```
patrickdevivo, line 20: is there a reason this needs to happen in `onChange` (vs when the Add button is clicked)? It seems possibly a bit expensive - I noticed on my machine it was a bit slow typing in the textarea - not sure if you see the same thing?

german-mergestat: I left this cuz it suggested me that, but now that you said it, confirm what I noticed as well, so.. I will remove it instead. let me take care of it then

patrickdevivo:
One thing I'm thinking about - it looks like this makes N GraphQL requests, one for each repo, right? Maybe we ought to have a "bulk add" mutation for adding a batch of repos

german-mergestat:
> 

Yes Patric we should to, something like passing just a repo string list from Frontend en then it create repos in one request. How can we address it? we have to implement a Graphile makeExtendSchemaPlugin that adds a mutation to create those repos. so... What should we do? address it in another issue to improve this part or implement now in this PR?

PR #175, opened 2022-08-31
feat: adding disabled column in repo_syncs & adding bool logic

ramiroJCB:
Here we add  a boolean  column in repo_syncs and we also  add logic that handles when to sync

### DIFF
--- a/internal/db/queries.sql
+++ b/internal/db/queries.sql
@@ -81,7 +81,7 @@ UPDATE mergestat.repo_sync_queue SET last_keep_alive = now() WHERE id = $1;
 
 -- name: EnqueueAllCompletedSyncs :exec
 INSERT INTO mergestat.repo_sync_queue (repo_sync_id, status)
-SELECT id, 'QUEUED' FROM mergestat.repo_syncs WHERE id NOT IN (SELECT repo_sync_id FROM mergestat.repo_sync_queue WHERE status = 'RUNNING' OR status = 'QUEUED')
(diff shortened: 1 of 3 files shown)


internal/db/queries.sql
```diff
@@ -81,7 +81,7 @@ UPDATE mergestat.repo_sync_queue SET last_keep_alive = now() WHERE id = $1;
 
 -- name: EnqueueAllCompletedSyncs :exec
 INSERT INTO mergestat.repo_sync_queue (repo_sync_id, status)
-SELECT id, 'QUEUED' FROM mergestat.repo_syncs WHERE id NOT IN (SELECT repo_sync_id FROM mergestat.repo_sync_queue WHERE status = 'RUNNING' OR status = 'QUEUED')
+SELECT id, 'QUEUED' FROM mergestat.repo_syncs WHERE enabled=true AND id NOT IN (SELECT repo_sync_id FROM mergestat.repo_sync_queue WHERE status = 'RUNNING' OR status = 'QUEUED')
```
patrickdevivo, line 84:
```suggestion
SELECT id, 'QUEUED' FROM mergestat.repo_syncs WHERE enabled AND id NOT IN (SELECT repo_sync_id FROM mergestat.repo_sync_queue WHERE status = 'RUNNING' OR status = 'QUEUED')
```

just a nit 😃, feel free to ignore. `enabled` is already a boolean so we don't need to do the comparison here

ramiroJCB: I like it !

ramiroJCB: lets do it

patrickdevivo: just remember to regenerate with `sqlc` as well 😃

ramiroJCB: Done :)

PR #184, opened 2022-09-02
feat: implement a `TRIVY_REPO_SCAN` sync type

patrickdevivo:
Sync type that runs `trivy repo ... -f json` and stashes the JSON output in the DB. Also includes a view to help extract common fields from the resulting JSON.

In general this is not a scalable way for us to continue adding sync types. However, for our current purposes this is sufficient to demonstrate the _types_ of syncs we want to continue building support for. We will soon need a better architecture and pattern for supporting arbitrary "3rd party" data sources such as `trivy` in a scalable and flexible way. That design is ongoing.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,10 @@ COPY migrations migrations
 
 COPY scripts/docker-init-entrypoint.sh docker-init-entrypoint.sh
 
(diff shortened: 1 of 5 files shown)


amenowanna (approved):
Love this so much! Great way to showcase the power of MergeStat and Fuse.

PR #187, opened 2022-09-03
feat(auto-import): Implementing auto import by user and organization

### LINKED ISSUE #103: Implement flow for creating a new auto-import
image

@german-mergestat see the Figma - reach out to me and I can point to the spots in it

german-mergestat:
Implementing auto import by user and organization. (Resolves #103)

1. Adding a General Context.
2. Adding mutation for registering auto import.
3. Adding Unit testing.

auto-import-user

auto-imported


ui/src/views/repositories/index.tsx
```diff
@@ -1,33 +1,49 @@
...
       <div className="bg-white border-b border-gray-200">
         <PageHeader />
         {showTable && <FilterHeader />}
       </div>
       <div className="flex-1 items-center p-8 overflow-auto">
+        {autoImportingRepos &&
+          <Alert type="info" theme="light" title="Auto importing repos" className='mb-5'>
+            The repos from the auto import setting will appear shortly once synced.
+          </Alert>
+        }
```
patrickdevivo, line 45: instead of this appearing only when the user creates a new auto-import, can we have it appear based on the result of a graphql query? We want it to show up when an auto-import _exists_ in the database, but the `last_import` timestamp is `NULL`. Does that make sense? That way this will persist across page refreshes as well

ui/src/views/repositories/index.tsx
```diff
@@ -1,33 +1,49 @@
...
   return (
     <main className="w-full flex flex-col h-full bg-gray-50 overflow-hidden">
       <div className="bg-white border-b border-gray-200">
         <PageHeader />
         {showTable && <FilterHeader />}
       </div>
       <div className="flex-1 items-center p-8 overflow-auto">
+        {autoImportingRepos &&
+          <Alert type="info" theme="light" title="Auto importing repos" className='mb-5'>
+            The repos from the auto import setting will appear shortly once synced.
```
patrickdevivo, line 43:
```suggestion
            Repositories from an auto-import will appear here once they are finished syncing.
```

ui/src/@types.tsx
```diff
@@ -32,7 +32,7 @@ export type RepoDataPropsT = {
   name: string
   createdAt: Date
   type: RepoType
-  automaticImport?: boolean
+  autoImportFrom?: string
```
JAORMX, line 35: Is this API v1 already? If so, you might want to use a different field or establish some migration path as this would be a breaking change.

german-mergestat: mmmm understand.. but if am not wrong the purpose of previous field `automaticImport` is the same that the new one `autoImportFrom`, (know if repo was auto imported or not), in this case with this new one, if `autoImportFrom` is `null`, then repo was not auto imported. That could be make sense? @patrickdevivo might you confirm please?

ui/src/views/repositories/index.tsx
german-mergestat: Done [fbbab8e](https://github.com/mergestat/fuse/pull/187/commits/fbbab8e107f753adfa46f0242be0c47ef8aeef57)

ui/src/views/repositories/index.tsx
german-mergestat: Done [fbbab8e](https://github.com/mergestat/fuse/pull/187/commits/fbbab8e107f753adfa46f0242be0c47ef8aeef57)

ui/src/@types.tsx
patrickdevivo: nothing is really "v1" yet as we're still pre-v1 in and under active development. This type represents a purely internal "api" for now as well, so it should be fine to change the type + name like so, only the UI is using it at the time being.

you are right though @JAORMX that in general and going forward "after v1" we should be careful about breaking changes to external facing fields (types and names) to minimize impact to downstream users/integrations and to communicate about it effectively. Thanks for keeping an eye out!

patrickdevivo (approved):

PR #237, opened 2022-09-09
feat(list-all-syncs): Listing all available sync types for any repo

### LINKED ISSUE #163: Repo sync page should list *all* available syncs
If a repo has no syncs associated with it, the `/repos/{id}` page looks like this: 
image

which isn't super helpful - instead, we should have all available sync types listed so the user can "flip" them on.

We should also derive the repo sync description and short name from the database

german-mergestat:
Listing all available sync types for any repo. (Resolves #163)

- Adding new mutation to record a new `sync type` for a repo
- Showing `short name` in sync types
- Disabling going to `/repos/${repository}/${syncTypeId}` when it is first time to sync
- Unit testing

list-sync-types

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -50,7 +50,7 @@
         "autoprefixer": "^10.4.4",
         "cross-env": "^7.0.3",
         "eslint": "^8.21.0",
-        "eslint-config-next": "^12.2.4",
+        "eslint-config-next": "^12.3.0",
         "eslint-config-prettier": "^8.3.0",
         "eslint-config-standard": "^17.0.0",
         "eslint-import-resolver-typescript": "^2.5.0",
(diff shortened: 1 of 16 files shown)


patrickdevivo:
Can we come up with an icon to use for when there's no sync history? We can chat tomorrow - I think I'd be fine with just the clock queued icon, unless there's a better "empty state" option

image

german-mergestat:
> Can we come up with an icon to use for when there's no sync history? We can chat tomorrow - I think I'd be fine with just the clock queued icon, unless there's a better "empty state" option
> 

No so convinced of use the same clock icon (due it already represent a queued sync). Maybe we could use this 
stroke. I see these [icons](https://www.figma.com/file/SGcDSC5dr5VxYSFP4x8p6i/MergeStat---Library?node-id=14%3A764) in figma, we can choose of them. Perhaps @emilyvandewalle could help us with this selection?

patrickdevivo:
what about a simple `-` to show it's kind of a N/A state? (https://heroicons.com/)

image

emilyvandewalle:
Indeed, I think a dash `-` to indicate empty could work (for example also used in Google Drive)

image

A blank circle is sometimes also used or an indeterminate icon could perhaps also work, but I think the dash would be the clearest 
image

patrickdevivo:
👍 I'm good with dash - let's do that

patrickdevivo (approved):

Issue #197, opened 2022-09-07
Implement basic authentication mechanism using Graphile JWT support


patrickdevivo:
- [x] Add the `roles` in postgres via a migration that graphile will use
- [x] Update local Graphile config to use a JWT
- [x] Implement an endpoint or mutation to allow generating a JWT
- [x] Implement a UI for authenticating via a secret
- [x] Update the helm chart to support auth setup

PR #243, opened 2022-09-12
Begin adding basic auth to the UI and GraphQL API

patrickdevivo:
Starts to address #197 

- Adds 2 database roles via a migration: `mergestat_anonymous` and `mergestat_admin`
- Updates the local docker-compose setup to make use of the roles and JWT based auth
- Adds an API endpoint (in the UI) at `/api/admin-auth` to authenticate a user who has a database user with a password

**NOTE**: in the current state of this PR, the UI will break as authentication is not handled anywhere. I suggest we keep this PR/branch open (I will keep it up to date) and eventually merge the UI changes we need in here, so that everything may be merged at once.


amenowanna:
@patrickdevivo can we please mark this PR as draft since it is not ready to merge.

ramiroJCB (approved):

german-mergestat (approved):

PR #247, opened 2022-09-12
feat(importer): added removeDeletedRepos flag in  autoimport settings

ramiroJCB:


internal/repos/importer.go
```diff
@@ -169,3 +183,22 @@ func (i *importer) handleGitHubImport(ctx context.Context, imp db.ListRepoImport
 
 	return tx.Commit(ctx)
 }
+
+func handleDeletedRepos(ctx context.Context, tx pgx.Tx, i *importer, repos []*githubRepo, repoOwner string) error {
+	var currentRepos []string
+	var err error
+
+	i.logger.Info().Msg("Starting to delete removed repos")
```
patrickdevivo, line 191:
```suggestion
	i.logger.Info().Msg("starting to delete removed repos")
```

(lower case just to be consistent for now)

internal/repos/importer.go
```diff
@@ -169,3 +183,22 @@ func (i *importer) handleGitHubImport(ctx context.Context, imp db.ListRepoImport
...
+
+	for _, repo := range repos {
+		currentRepos = append(currentRepos, fmt.Sprintf("https://github.com/%s/%s", repoOwner, repo.Name))
+	}
+
+	if err = i.db.WithTx(tx).DeleteRemovedRepos(ctx, currentRepos); err != nil {
+		return err
+	}
+
+	i.logger.Info().Msg("Deletion of removed repos completed")
```
patrickdevivo, line 201:
```suggestion
	i.logger.Info().Msg("deletion of removed repos completed")
```

same as above

internal/repos/importer.go
```diff
@@ -169,3 +183,22 @@ func (i *importer) handleGitHubImport(ctx context.Context, imp db.ListRepoImport
 
 	return tx.Commit(ctx)
 }
+
+func handleDeletedRepos(ctx context.Context, tx pgx.Tx, i *importer, repos []*githubRepo, repoOwner string) error {
```
patrickdevivo, line 187:
```suggestion
func (i *importer) handleDeletedRepos(ctx context.Context, tx pgx.Tx, repos []*githubRepo, repoOwner string) error {
```

might as well make it a "private method" on the importer? One less arg to pass around

internal/repos/importer.go
```diff
@@ -141,6 +150,11 @@ func (i *importer) handleGitHubImport(ctx context.Context, imp db.ListRepoImport
 		}
 	}()
 
+	if removeDeletedRepos {
+		err = handleDeletedRepos(ctx, tx, i, repos, repoOwner)
+		return err
```
patrickdevivo, line 155: we should probably check if this error here before the `return` I'm guessing?

internal/repos/importer.go
```diff
@@ -127,12 +134,14 @@ func (i *importer) handleGitHubImport(ctx context.Context, imp db.ListRepoImport
 		return err
 	}
 
-	// start a new pg transaction to handle the repo upsert
+	// start a new pg transaction to handle the repo upsert  and handle deletion of removed repos
```
patrickdevivo, line 137:
```suggestion
	// start a new pg transaction to handle the repo upsert and handle deletion of removed repos
```

just removing an extra space

internal/db/queries.sql
```diff
@@ -96,3 +96,7 @@ INSERT INTO mergestat.repo_sync_logs (repo_sync_queue_id, log_type, message)
 SELECT id, 'ERROR', 'No response from job within reasonable interval. Timing out.' FROM timed_out_sync_jobs
 RETURNING repo_sync_queue_id
 ;
+
+-- name: DeleteRemovedRepos :exec 
+DELETE FROM public.repos WHERE repo NOT IN (@currentRepos::string[]);
```
patrickdevivo, line 101:
```suggestion
DELETE FROM public.repos WHERE NOT (repo = ANY (@currentRepos::text[]));
```

not totally sure (will need to test locally) but try this?

internal/db/queries.sql
```diff
@@ -96,3 +96,8 @@ INSERT INTO mergestat.repo_sync_logs (repo_sync_queue_id, log_type, message)
 SELECT id, 'ERROR', 'No response from job within reasonable interval. Timing out.' FROM timed_out_sync_jobs
 RETURNING repo_sync_queue_id
 ;
+
+-- name: DeleteRemovedRepos :exec 
+DELETE FROM public.repos WHERE repo NOT IN (@currentRepos::text[])
```
amenowanna, line 101: I started a review yesterday but cancelled it once I saw @patrickdevivo had started one too. The only comment I have is that I think we should include the context of the `repo_import_id` in the where clause here to only compare against the repos imported by that import. This will allow other imports to continue working as well as manually added repos.

internal/repos/importer.go
```diff
@@ -141,6 +150,14 @@ func (i *importer) handleGitHubImport(ctx context.Context, imp db.ListRepoImport
 		}
 	}()
 
+	if removeDeletedRepos {
+		err = i.handleDeletedRepos(ctx, tx, repos, repoOwner, imp.ID)
+
```
patrickdevivo, line 155:
```suggestion
```

patrickdevivo (dismissed):
was able to validate locally,  working as expected from what I can tell!

internal/repos/importer.go
ramiroJCB: Makes sense

patrickdevivo (approved):

PR #256, opened 2022-09-14
feat(remove-repo): Implementing functionality to remove a repo

### LINKED ISSUE #101: Implement remove repo button/flow from repos page
Implement the button to remove (delete) a repository (via a graphql mutation).

We should also include a confirmation modal to add another step for the user, since this will be a destructive action (will lose all data for this repo as well)

image

german-mergestat:
Implementing functionality to remove a repo. (Resolves #101)

- Show card `Start by adding a GitHub authentication token` when there's no repos.
- It was creating several queries with polling interval. This bug was fixed.
- No Unit testing included. I will include the unit testing in next issue which has an easier `button` element to grab.

remove-repo

### DIFF
--- a/ui/src/views/repositories/modals/remove-repository-modal/index.tsx
+++ b/ui/src/views/repositories/modals/remove-repository-modal/index.tsx
@@ -1,5 +1,5 @@
 import { ApolloError, useMutation } from '@apollo/client'
-import { Button, Modal, Toolbar } from '@mergestat/blocks'
+import { Alert, Button, Modal, Toolbar } from '@mergestat/blocks'
 import { TrashIcon, XIcon } from '@mergestat/icons'
 import React, { useCallback } from 'react'
 import REMOVE_REPO from 'src/api-logic/graphql/mutations/remove-repo'
@@ -48,7 +48,12 @@ export const RemoveRepositoryModal: React.FC = () => {
         </Toolbar>
       </Modal.Header>
       <Modal.Body>
-        <div className="p-6">¿Are you sure to remove <b>{repoToRemove?.name}</b>?</div>
+        {repoToRemove?.autoImported &&
+          <Alert type="warning" theme="light" className='mx-4 mt-6 mb-2 text-xs'>
+            This repo was added by an automatic import, it will be re-added on the next run, unless the repo auto import is removed.
+          </Alert>
+        }
+        <div className="px-6 pt-2 pb-6">Are you sure you want to remove <b>{repoToRemove?.name}</b>?</div>
       </Modal.Body>
       <Modal.Footer>
         <Toolbar className="h-16 px-6">
@@ -69,7 +74,7 @@ export const RemoveRepositoryModal: React.FC = () => {
                 startIcon={<TrashIcon className="t-icon" />}
                 className="my-3"
               >
-                Delete Repository
+                Remove Repository
               </Button>
             </Toolbar.Item>
           </Toolbar.Right>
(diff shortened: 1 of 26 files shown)


ui/src/views/repositories/modals/remove-repository-modal/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+                skin="borderless-muted"
+                onClick={close}
+                startIcon={<XIcon className="t-icon" />}
+              />
+            </Toolbar.Item>
+          </Toolbar.Right>
+        </Toolbar>
+      </Modal.Header>
+      <Modal.Body>
+        <div className="p-6">¿Are you sure to remove <b>{repoToRemove?.name}</b>?</div>
```
amenowanna, line 51: The Spanish ¿ came out!

patrickdevivo:
```suggestion
        <div className="p-6">Are you sure to remove <b>{repoToRemove?.name}</b>?</div>
```

ui/src/views/repositories/modals/remove-repository-modal/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+              </Button>
+              <Button
+                skin="danger"
+                onClick={() => removeRepo({
+                  variables: { id: repoToRemove?.id }
+                })}
+                startIcon={<TrashIcon className="t-icon" />}
+                className="my-3"
+              >
+                Delete Repository
```
patrickdevivo, line 72:
```suggestion
                Remove Repository
```

For consistency, let's use `Remove` when referring to this action everywhere in the UI. I worry `Delete` sounds a little strong and may incorrectly imply we're deleting the repo from GitHub, which definitely isn't the case. Remove sounds a little softer IMO and clearer that we're just removing it from MergeStat

ui/src/views/repositories/modals/remove-repository-modal/index.tsx
patrickdevivo: One other request 😃 - if the repo has an `auto_import_id` associated with it (i.e. it came from a repo auto import) can we have an additional message here, in a `warning` callout that says "This repo was added by an automatic import, it will be re-added on the next run, unless the repo auto import is removed."

amenowanna: Are you sure you want to make that suggestion?

patrickdevivo:
This is an awesome PR! 🎉 

However, this card should _only_ be shown if there are no repos **and** there's not a GitHub PAT set (`service_auth_credentials` of type `GITHUB_PAT` in the DB does not exist). If there is one set, we should just show a default "no repos" message for now. We have in Figma a version of that (see below)

image

image

ui/src/views/repositories/modals/remove-repository-modal/index.tsx
german-mergestat: Done [bc53f39](https://github.com/mergestat/fuse/pull/256/commits/bc53f3927b43a41ca7584735026bf686c432609a)

ui/src/views/repositories/modals/remove-repository-modal/index.tsx
german-mergestat: Done [bc53f39](https://github.com/mergestat/fuse/pull/256/commits/bc53f3927b43a41ca7584735026bf686c432609a)

patrickdevivo (approved):

PR #260, opened 2022-09-15
fix: add a `nil` pointer check and some warn logging

patrickdevivo:
in the newly added `GIT_BLAME` sync. This is a "hotfix" and we should address the underlying issue in the library we're using at some point.

### DIFF
--- a/internal/syncer/git_blame.go
+++ b/internal/syncer/git_blame.go
@@ -127,10 +127,26 @@ func (w *worker) handleGitBlame(ctx context.Context, j *db.DequeueSyncJobRow) er
 
 		for lineIdx, blame := range res {
 			lineNo := lineIdx + 1
+
+			if blame == nil {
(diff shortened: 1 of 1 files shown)


internal/syncer/git_blame.go
```diff
@@ -127,10 +127,26 @@ func (w *worker) handleGitBlame(ctx context.Context, j *db.DequeueSyncJobRow) er
...
+			var authorEmail, authorName *string
+			var authorWhen *time.Time
+			// TODO(patrickdevivo) we shouldn't be seeing a nil Author here, but we are
+			// until we can audit what's going on in the `gitutils` package let's add a check here
+			if blame.Author != nil {
+				authorEmail = &blame.Author.Email
+				authorName = &blame.Author.Name
+				authorWhen = &blame.Author.When
+			}
+
```
ramiroJCB, line 145: Interesting solution

ramiroJCB (approved):

patrickdevivo:
seems to be result of this issue: https://github.com/mergestat/gitutils/issues/5 where we return a `nil` blame line when we shouldn't (on empty files)

PR #264, opened 2022-09-15
feat(dockerfile): Added  cache option when building image

ramiroJCB:
Please @patrickdevivo  @amenowanna   test this on your local env :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,5 @@
+# syntax=docker/dockerfile:1-experimental
+
 FROM golang:1.18-buster AS builder
 RUN set -x && apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y ca-certificates cmake libssl-dev
 COPY scripts/install_libgit2.sh install_libgit2.sh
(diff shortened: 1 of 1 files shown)


amenowanna:
Man...I tried this before and it didn't seem to work. But this seems to be working! Made a change in main.go and the build time was 8.9 seconds!

ramiroJCB:
Will do  a rebase after https://github.com/mergestat/fuse/pull/266 is merge

migrations/20220907164245_imported_at_publics.up.sql
```diff
@@ -79,7 +79,6 @@ CREATE OR REPLACE VIEW trivy_repo_vulnerabilities AS
 SELECT
     trivy_repo_scans.repo_id,
     v as vulnerability,
-    trivy_repo_scans._mergestat_synced_at,
```
patrickdevivo, line 82: let's remove this

Dockerfile
```diff
@@ -6,7 +8,8 @@ WORKDIR /src
 COPY go.mod go.sum ./
 RUN go mod download && go mod verify
 COPY . .
-RUN PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/usr/share/pkgconfig/libgit2/lib/pkgconfig/ make
+RUN  --mount=type=cache,target=/root/.cache/go-build \
```
patrickdevivo, line 11:
```suggestion
RUN --mount=type=cache,target=/root/.cache/go-build \
```

rm extra space

patrickdevivo (approved):
awesome! really excited about this one - should save us a lot of time waiting around 😄 

https://pkg.go.dev/cmd/go#hdr-Build_and_test_caching

patrickdevivo:
```
The build cache correctly accounts for changes to Go source files, compilers, compiler options, and so on: cleaning the cache explicitly should not be necessary in typical use. However, the build cache does not detect changes to C libraries imported with cgo. If you have made changes to the C libraries on your system, you will need to clean the cache explicitly or else use the -a build flag (see 'go help build') to force rebuilding of packages that depend on the updated C libraries.
```

this may be something to be aware of generally, in case we bump a libgit2 or sqlite version

PR #266, opened 2022-09-15
chore(migrations): fixing error introduced by sqlfluff

### LINKED ISSUE #267: migration bug(potentially introduced by sqlfluff)
We think this issue was introduced by sqlfluff automatic fix .
If that is the case we need to be carefull  when using it and re-test every query formatted by sqlfluff in the future (is we keep it )

Update: We now know that qlfluff moved trivy_repo_scans.mergestat_synced_at to line 82 creating the migration bug , because that's the key that I added on that migration  so by moving it back to the end solves the issue :slightly_smiling_face:

Image

Image

ramiroJCB:
Closes #267

### DIFF
--- a/migrations/20220907164245_imported_at_publics.up.sql
+++ b/migrations/20220907164245_imported_at_publics.up.sql
@@ -76,7 +76,7 @@ FROM public.git_refs
 WHERE (git_refs.type = 'tag'::text);
 
 CREATE OR REPLACE VIEW trivy_repo_vulnerabilities AS
-SELECT
+SELECT -- noqa: L034
     trivy_repo_scans.repo_id,
     v as vulnerability,
     r.value->> 'Target' as target,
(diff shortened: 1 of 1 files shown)


patrickdevivo:
ah I am understanding the issue more now - very interesting. so the linter wants us to change the order of the columns in the view, but postgres doesn't like that (i think?) because it thinks the schema of the view is changing, even tho it's just a column re-ordering

migrations/20220907164245_imported_at_publics.up.sql
```diff
@@ -79,7 +79,6 @@ CREATE OR REPLACE VIEW trivy_repo_vulnerabilities AS
 SELECT
```
patrickdevivo, line 79:
```suggestion
SELECT -- noqa: L034
```

patrickdevivo (approved):

ramiroJCB:
> ah I am understanding the issue more now - very interesting. so the linter wants us to change the order of the columns in the view, but postgres doesn't like that (i think?) because it thinks the schema of the view is changing, even tho it's just a column re-ordering

is because of this. 

> `CREATE OR REPLACE VIEW is similar, but if a view of the same name already exists, it is replaced. The new query must generate the same columns that were generated by the existing view query (that is, the same column names in the same order and with the same data types), but it may add additional columns to the end of the list. The calculations giving rise to the output columns may be completely different.`

PR #279, opened 2022-09-16
feat(migrations): Adding PK to git commit stats

ramiroJCB:


migrations/20220916222411_pk_commit_stats.up.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE public.git_commit_stats
+ADD CONSTRAINT commit_stats_pkey PRIMARY KEY(file_path,commit_hash);
```
amenowanna, line 4: should include repo_id to account for forked repos

patrickdevivo (approved):

PR #298, opened 2022-09-19
feat(worker): Adding priority logic for sync queues

ramiroJCB:


migrations/20220919161227_priority_column_queue.up.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+ALTER TABLE mergestat.repo_sync_queue
+ADD COLUMN priority_level NOT NULL INTEGER DEFAULT 0;
```
patrickdevivo, line 4:
```suggestion
ADD COLUMN priority NOT NULL INTEGER DEFAULT 0;
```
just a suggestion - for me, `priority_level` implies discrete levels like `low` `medium` `high` vs just `priority` doesn't have that same implication, and seems like it fits the numeric approach we're taking here

internal/db/queries.sql
```diff
@@ -34,7 +34,7 @@ WITH dequeued AS (
 	WHERE id IN (
 		SELECT id FROM mergestat.repo_sync_queue
 		WHERE status = 'QUEUED'
-		ORDER BY repo_sync_queue.created_at ASC, repo_sync_queue.id ASC LIMIT 1 FOR UPDATE SKIP LOCKED
+		ORDER BY repo_sync_queue.priority_level ASC,repo_sync_queue.created_at ASC, repo_sync_queue.id ASC LIMIT 1 FOR UPDATE SKIP LOCKED
```
patrickdevivo, line 37:
```suggestion
		ORDER BY repo_sync_queue.priority_level DESC, repo_sync_queue.created_at ASC, repo_sync_queue.id ASC LIMIT 1 FOR UPDATE SKIP LOCKED
```

we'll want higher priorities to be dequeued first (I assume), so I think we want this DESC

amenowanna: Ascending seems right to me priority 1 goes before priority 2.

patrickdevivo: You're right - that makes more sense to me too

internal/db/queries.sql
```diff
@@ -33,8 +33,8 @@ WITH dequeued AS (
 	UPDATE mergestat.repo_sync_queue SET status = 'RUNNING'
 	WHERE id IN (
 		SELECT id FROM mergestat.repo_sync_queue
-		WHERE status = 'QUEUED'
-		ORDER BY repo_sync_queue.created_at ASC, repo_sync_queue.id ASC LIMIT 1 FOR UPDATE SKIP LOCKED
+		WHERE status = 'QUEUED' AND priority >=0
```
patrickdevivo, line 36:
```suggestion
		WHERE status = 'QUEUED'
```

I think we're okay to allow for negative priorities. If we don't, I think that should be a constraint on that column in the table for values going in, vs handling it here. But I don't see an issue (in fact it may be important) to allow for an "infinite upper bound" on priority in case we ever need to _ensure_ a job happened before all others

internal/db/queries.sql
```diff
@@ -33,8 +33,8 @@ WITH dequeued AS (
...
+		ORDER BY repo_sync_queue.priority DESC,repo_sync_queue.created_at ASC, repo_sync_queue.id ASC LIMIT 1 FOR UPDATE SKIP LOCKED
```
patrickdevivo, line 37:
```suggestion
		ORDER BY repo_sync_queue.priority DESC, repo_sync_queue.created_at ASC, repo_sync_queue.id ASC LIMIT 1 FOR UPDATE SKIP LOCKED
```

just adding a space here

internal/db/queries.sql
```diff
@@ -34,7 +34,7 @@ WITH dequeued AS (
 	WHERE id IN (
 		SELECT id FROM mergestat.repo_sync_queue
 		WHERE status = 'QUEUED'
-		ORDER BY repo_sync_queue.created_at ASC, repo_sync_queue.id ASC LIMIT 1 FOR UPDATE SKIP LOCKED
+		ORDER BY repo_sync_queue.priority DESC, repo_sync_queue.created_at ASC, repo_sync_queue.id ASC LIMIT 1 FOR UPDATE SKIP LOCKED
```
patrickdevivo, line 37:
```suggestion
		ORDER BY repo_sync_queue.priority ASC, repo_sync_queue.created_at ASC, repo_sync_queue.id ASC LIMIT 1 FOR UPDATE SKIP LOCKED
```

patrickdevivo: @ramiroJCB just FYI

patrickdevivo (approved):

Issue #198, opened 2022-09-07
Implement a simple login UI + functionality


patrickdevivo:
As this ticket progresses: https://github.com/mergestat/fuse/issues/197 we will want to also work on a basic login flow in the UI. This includes the design of a simplified login page based on the screenshot here, and the front-end code implementation to begin passing a JWT supplied by the user (once authenticated).

<img width="2048" alt="image" src="https://user-images.githubusercontent.com/57259/188934646-f9dd3dca-02ff-4fa0-8d1c-526b09a9196d.png">

- [x] Implement basic login UI in Figma 
- [x] Implement basic login UI in code
- [x] Connect login UI to new authentication behavior in front-end

emilyvandewalle:
The first Figma designs for this can be found [here](https://www.figma.com/file/a6uVmHun0vKY08mGAsvHjJ/MergeStat---Self-Hosted-App?node-id=3063%3A58011)

## Log in page
![image](https://user-images.githubusercontent.com/36261498/191312070-957cae57-61b5-44ee-a833-2147a51f103a.png)


## Logged out state
![image](https://user-images.githubusercontent.com/36261498/191313228-7bb85d65-95c2-424a-8464-2073fb2c066b.png)


- More of a side detail, but while I was at it I wanted to experiment with updating the banner/alert style a bit. The current colors feel a bit harsh and are based on the default Tailwind colors (which use different hues for the different tints). I've experimented with making them a bit softer, which I think could also improve the readability. What do you guys think?

## Error state
![image](https://user-images.githubusercontent.com/36261498/191314019-fa452660-444c-440a-9b2a-9c41f2d2706c.png)

If this looks good, I'll add this layout in the `fuse` repo

PR #305, opened 2022-09-21
feat: add login ui

emilyvandewalle:
(Related to #198)

## Login UI
- Added the UI for the login screen
- Added dummy variables for the `loggedOut` and `error` states

**Default**
image

**Logged out**
image

**Error**
image

## NavHeader

- I added the logged in username to the `NavHeader` component and a logout link that redirects to the authentication page to keep it simple for now. In the future we could reintroduce the user menu again
- I also updated the version tag a bit to make it look more interactive

image


amenowanna:
This looks great. The only feedback I have is that we should use the term either `role` or `user` instead of name since we are using the database primitives for this.

amenowanna:
Apologies for closing the PR...that was not intentional.

german-mergestat (dismissed):
ok, this is just the scaffolding, LGTM 🙂

german-mergestat (approved):

Issue #311, opened 2022-09-22
Deleting an `auto_import` should also delete the repos it imported


patrickdevivo:
We can do this by adding a `CASCADE` to the `repo_import_id` foreign key constraint on the `repos` table, similar to what was done for the data tables pointing back to `repos`.

This will enable #309 (it will be important to have that warning in the UI)

ramiroJCB:
Should we tackle this ticket faster ? I should be able to do in a few minutes @patrickdevivo

patrickdevivo:
yep - go for it

PR #314, opened 2022-09-22
feat(migrations): CASCADE for all repos when import is deleted

ramiroJCB:
Reference :) https://github.com/mergestat/fuse/issues/311


ramiroJCB:
Tested locally  and it works :)

patrickdevivo (approved):

PR #313, opened 2022-09-22
feat(worker): Update enqueue system to not starve low prio queues

ramiroJCB:
Direct link to the issue https://github.com/mergestat/fuse/issues/318 :)

PR #317, opened 2022-09-23
Default syncs

### LINKED ISSUE #78: Investigate how to approach "default" sync types for repos
When a new repo is added, either manually or via an auto-import, it won't have any syncs enabled for it. This can be a frustrating user experience for anyone who deals with many repos (many new repos). We should investigate how to define a "default" set of syncs that are enabled for a repo when first added, so that syncing automatically begins when a new repo is added.

ramiroJCB:
In this code we open the possibility for users  to have default sync types for importing repos , we have had issues in the past with users that  have over 500> repos to import and have to manually set  the syncs for each one .
We needed to do it without duplicating syncs for repos  or re-enqueue  syncs that were already queued  .
One way to test this implementation is to add an import from the db with defeault syncs types 
Next query can do the job 
```
INSERT INTO repo_imports (type, settings)
VALUES ('GITHUB_USER', '{"user": "ramiroJCB","defaultSyncTypes":["GIT_FILES","GIT_COMMITS","GIT_REFS"]}');

```
This implementation does not cover single imports from github  , so this only works from imports ,so if you go and click on 'ADD REPOSITORIES' in the ui the repository added by that option wont have default syncs applied  to it .
Screen Shot 2022-09-23 at 11 51 46 AM

Closes #78


internal/repos/importer.go
```diff
@@ -204,3 +215,54 @@ func (i *importer) handleDeletedRepos(ctx context.Context, tx pgx.Tx, repos []*g
...
+
+	return err
+}
+
+func (i *importer) insertSyncInQueue(ctx context.Context, tx pgx.Tx, defaultSyncs []string, repoID uuid.UUID) error {
+	var err error
+	var syncExistance bool
+	//testing methods that will be removed after draft
+	defaultSyncs = append(defaultSyncs, "GIT_COMMITS")
+	defaultSyncs = append(defaultSyncs, "GIT_COMMIT_STATS")
```
ramiroJCB, line 248: @patrickdevivo  @amenowanna   guys please this line will be the line test  locally  the default syncs we want , will be removed after draft phase  along with the addition of the if statement for these logic to kick off

internal/repos/importer.go
```diff
@@ -84,13 +84,15 @@ var (
 )
 
 type githubOrgRepoImportSettings struct {
-	Org                string `json:"org"`
-	RemoveDeletedRepos bool   `json:"removeDeletedRepos"`
+	Org                string   `json:"org"`
+	RemoveDeletedRepos bool     `json:"removeDeletedRepos"`
+	DefaultSyncs       []string `json:"defaultSyncs"`
```
amenowanna, line 89: @ramiroJCB could you please share an example config JSON for this. Are you going to be using repo_sync_types here? If so, I think we should call it types here as well.

ramiroJCB: Yeah , my suggestion was that we show the user in a way our available syncs types to them and the pick.
So yeah  you like **DefaultSyncTypes**  or just **SyncTypes**

amenowanna: DefaultSyncTypes makes sense to me.

cmd/worker/main.go
```diff
@@ -219,7 +219,7 @@ func main() {
 	// either via the database/app or possibly with env vars
 	go func() {
 		defer wg.Done()
-		scheduler.New(&logger, pool).Start(ctx, 10*time.Minute)
+		scheduler.New(&logger, pool).Start(ctx, 1*time.Minute)
```
patrickdevivo, line 222: This is an extraneous change from the other PR I think (#313)

internal/repos/importer.go
```diff
@@ -204,3 +215,54 @@ func (i *importer) handleDeletedRepos(ctx context.Context, tx pgx.Tx, repos []*g
...
+		return err
+	}
+
+	for _, repoID := range repoIDs {
+		if err = i.insertSyncInQueue(ctx, tx, defaultSyncs, repoID); err != nil {
+			i.logger.Err(err)
+			return err
+		}
+
+	}
```
patrickdevivo, line 236: IMO we should remove this functionality from here, if I'm understanding this correctly. It looks like you're also enqueuing the newly added `repo_sync`s to run during the import? To me, that should be a separate process, we'll let the scheduler pick those up and enqueue with everything else. I think the code here should just _enable_ the default sync types for a repo, but not trigger it to run

migrations/20220922230639_insert_default_sync_fn.up.sql
```diff
@@ -0,0 +1,16 @@
...
+$$
+begin
+IF EXISTS (SELECT sync_type FROM mergestat.repo_syncs WHERE repo_id = imported_repo_id_param AND sync_type=sync_type_param) THEN
+    RETURN TRUE;
+ELSE 
+    INSERT INTO mergestat.repo_syncs (repo_id, sync_type) VALUES(imported_repo_id_param,sync_type_param);
+    RETURN FALSE;
+END IF;
+end;
+$$ LANGUAGE plpgsql;
```
patrickdevivo, line 14: rather than defining a DB function for this, I think we ought to do it directly as a statement that `sqlc` uses (see above comment). I think it's straight forward enough to handle that way, and then we don't need a migration for this functionality

patrickdevivo (changes requested):

internal/db/queries.sql
```diff
@@ -103,3 +103,18 @@ DELETE FROM public.repos WHERE repo_import_id = $1::uuid AND NOT(repo = ANY($2::
 
 -- name: CleanOldRepoSyncQueue :exec
 SELECT mergestat.simple_repo_sync_queue_cleanup($1::INTEGER);
+
+-- name: GetAllReposId :many
+SELECT id FROM public.repos WHERE repo_import_id = @importID::uuid
+;
+
+-- name: AddingNewDefaultSync :one
+SELECT mergestat.insert_default_sync(@repoID::uuid,@syncType::text);
```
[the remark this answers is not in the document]
ramiroJCB: Yes , I have the wrong idea of enqueuing  the newly  added syncs so I needed  the boolean from the function. 
Now It can be a online insert with do nothing on conflict

migrations/20220922230639_insert_default_sync_fn.up.sql
ramiroJCB: Yeak  makes sense , I was making this change because of the enqueue misconception I had ;/

internal/repos/importer.go
```diff
@@ -177,6 +194,19 @@ func (i *importer) handleGitHubImport(ctx context.Context, imp db.ListRepoImport
...
 	}
 
+	//handle default syncs on new repo import
+	if defaultSyncs && len(newImportedRepoUrls) > 0 {
+		err = i.handleDefaultSyncs(ctx, tx, defaultSyncTypes, imp.ID, newImportedRepoUrls)
+		if err != nil {
+			return err
+		}
+
+		i.logger.Info().Msg("Enqueuing all available syncs for newly imported repos")
```
patrickdevivo, line 204:
```suggestion
		i.logger.Info().Msg("enqueuing all available syncs for newly imported repos")
```

let's be consistent

internal/repos/importer.go
```diff
@@ -204,3 +232,81 @@ func (i *importer) handleDeletedRepos(ctx context.Context, tx pgx.Tx, repos []*g
 
 	return err
 }
+
+func (i *importer) getNewImportedRepoUrls(ctx context.Context, tx pgx.Tx, repos []*githubRepo, repoOwner string, impID uuid.UUID) ([]string, error) {
```
patrickdevivo, line 236:
```suggestion
// getNewImportedRepoUrls retrieves the URLs of all repos that are now being _added_ by an import
func (i *importer) getNewImportedRepoUrls(ctx context.Context, tx pgx.Tx, repos []*githubRepo, repoOwner string, impID uuid.UUID) ([]string, error) {
```

internal/repos/importer.go
```diff
@@ -204,3 +232,81 @@ func (i *importer) handleDeletedRepos(ctx context.Context, tx pgx.Tx, repos []*g
...
+		}
+		if unique {
+			newImportedRepoUrls = append(newImportedRepoUrls, imported)
+		}
+	}
+
+	return newImportedRepoUrls, err
+}
+
+func (i *importer) handleDefaultSyncs(ctx context.Context, tx pgx.Tx, defaultSyncTypes []string, impID uuid.UUID, newlyAddedRepos []string) error {
```
patrickdevivo, line 270:
```suggestion
// handleDefaultSyncs if the repo auto import has any default syncs defined, insert those into the DB
func (i *importer) handleDefaultSyncs(ctx context.Context, tx pgx.Tx, defaultSyncTypes []string, impID uuid.UUID, newlyAddedRepos []string) error {
```

patrickdevivo (approved):

PR #321, opened 2022-09-24
feat: add `old_file_mode` and `new_file_mode` columns

### LINKED ISSUE #293: Re-evaluate primary key for `git_commit_stats` table
Creating a ticket to track an issue we're seeing where the PK added here: https://github.com/mergestat/fuse/pull/279 seems not "unique enough" for certain edge cases. Essentially, it's possible to have multiple changes (diff stats) for the same `path` in a commit diff. The example we encountered was with a regular file being remove at `some/path` and then a git submodule being added at the same `some/path`, causing a PK constraint error when inserting the stats for this commit.

Maybe we need to add/track another flag here that tracks the file type with the stats for it?

patrickdevivo:
to `git_commit_stats` table. This should allow us to add a primary key safely on (repo_id, file_path, commit_hash, new_file_mode) (at least I believe so, in my understanding of these columns). This addresses the conflict issue we saw for [this commit](https://github.com/mergestat/libgit2/commit/156cfec096fd71cbd6d15592e80d164b8f3b55ad) (`tests-clar/clar`) when trying to add a PK _without_ `new_file_mode`

We still need to figure out how to migrate existing contents of `git_commit_stats` to use the new primary key here without a conflict. I suppose the easiest would be to `TRUNCATE` the table (and be okay with the data loss) and allow users to re-sync? @amenowanna would love to know your suggestions here.

Closes #293


amenowanna:
We have discussed that with these changes and some manual prep this migration will work. In the future we will need to look at creating new tables and then migrating the data over as the data we store increases over time.

amenowanna (approved):

PR #322, opened 2022-09-24
feat: experiment using `@floating-ui` for tooltips

patrickdevivo:
seems like a better option than rolling our own solution from scratch.


patrickdevivo:
Closing in favor of this work: https://github.com/mergestat/blocks/issues/42

PR #327, opened 2022-09-27
fix: adding default value and NOT NULL

amenowanna:
PK is now allowed to have null values so we need to set a default value for the new columns and set them to NOT NULL


patrickdevivo (approved):

ramiroJCB (approved):

ramiroJCB:
Small improvement on the commit message will be  "fix(migrations):......"

Issue #88, opened 2022-07-29
Implement the UI for direct SQL queries


patrickdevivo:
Implement the UI for #87.

We should have sufficient designs in place from the `app.mergestat.com` implementation.

emilyvandewalle:
The Figma designs so far for this can be found [here](https://www.figma.com/file/a6uVmHun0vKY08mGAsvHjJ/MergeStat---Self-Hosted-App?node-id=3102%3A58384). It's still work in progress and I didn't get to some details yet for the future version (like the auto complete, database references, query charts, etc...), but wanted to check in between for feedback:

# First simple version

## Initial state
![image](https://user-images.githubusercontent.com/36261498/192336295-8f72443a-61a5-4c2f-939c-60d7af874a00.png)
- Experimented with moving the main actions (`Execute`) in the top header/toolbar (instead of part of the query editor)

## Loading state
![image](https://user-images.githubusercontent.com/36261498/192336602-f18ba858-ee09-4285-b5a7-ee6d0fd61199.png)
- Using a skeleton loading could decrease the perceived loading time and give the user a better expectation of the loaded state
- Mostly used for initial full-page loading states, but I think it could work here too
- **Reference:** I thought this was a good in-depth reference in general for loading feedback: [https://pencilandpaper.io/articles/ux-pattern-analysis-loading-feedback/#tldr-core-principles](https://pencilandpaper.io/articles/ux-pattern-analysis-loading-feedback/#tldr-core-principles ). This article recommends different loading approaches based on the loading time (`indeterminate` vs `determinate` for longer than 10s). Do we know an average loading time or does it vary a lot depending on the query/other factors?

## Success state
![image](https://user-images.githubusercontent.com/36261498/192336649-22623504-047d-4270-ae7a-d1cb78cc2427.png)
- Experimented with making the results table full width to make better use of the space available

### More results than row limit
![image](https://user-images.githubusercontent.com/36261498/192336779-0e632648-4bac-496c-b9f8-e565b53b19df.png)

### Fullscreen results view
![image](https://user-images.githubusercontent.com/36261498/192336894-9a81565e-0820-4671-a7e1-6b779584fbd9.png)
- Maybe we could make the `Query results` view expandable/full screen (a bit like in popsql) as I can imagine once you've ran the query sometimes you want to zoom in on the results. 
- We could also keep the resizing of the `Query editor` vs. `Query results` area which also helps with that

## Error state
[...]

PR #329, opened 2022-09-28
Add query editor and UI for executing ad-hoc SQL

emilyvandewalle:
Added the basic query UI with its different states:

I used a dummy variable again (`const state: States = States.Filled` which can be `Filled`, `Empty`, `Error` or `Loading`) to toggle between the different states

## Initial
image

## Loading
image

## Error
image

## Filled
image

(Related to #88)


ui/src/views/sql-query-editor/components/state-filled.tsx
```diff
@@ -0,0 +1,239 @@
...
+    }
+    return text
+  }
+
+  const copyToClipboard = () => {
+    copy(getText())
+  }
+
+  const exportData = () => {
+    const jsonString = `data:text/${exportFormat.toLowerCase()};chatset=utf-8,${encodeURIComponent(getText())}`
```
patrickdevivo, line 65:
```suggestion
    const jsonString = `data:text/${exportFormat.toLowerCase()};charset=utf-8,${encodeURIComponent(getText())}`
```

I'm guessing that's a typo?

patrickdevivo (changes requested):

patrickdevivo (approved):

PR #352, opened 2022-10-01
feat: add a spinner to the auto import alert

### LINKED ISSUE #345: Add animated spinner for repo auto import alert banner on `/repos` page
Just to make it clearer that something is happening in the background

patrickdevivo:
the message that's displayed when a repo auto import is running for the first time.

Closes #345 

image

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -10,7 +10,7 @@
       "dependencies": {
         "@apollo/client": "^3.6.9",
         "@apollo/react-testing": "^4.0.0",
-        "@mergestat/blocks": "^1.2.8",
+        "@mergestat/blocks": "^1.4.0",
         "@mergestat/icons": "^1.1.2",
         "@next/bundle-analyzer": "^12.1.4",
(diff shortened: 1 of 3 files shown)


emilyvandewalle:
The spinner is quite subtle on the right, maybe we could add a `Loader` variant to the `Alert` component, like this?

image

patrickdevivo:
@emilyvandewalle yep you're right - a `Loader` variant to the `Alert` component would be great here. In fact, I was thinking that the `Alert` component should let the icon be _any_ DOM element (or React component) vs relying the `variant` to drive that. So maybe the default is set according to the `variant`, but it can be overridden by an `icon` prop. What do you think @german-mergestat?

I will open a ticket in `block`

patrickdevivo:
This PR: https://github.com/mergestat/blocks/pull/78 should enabled this

patrickdevivo:
Latest commit refactors to look like this:

image

@emilyvandewalle @german-mergestat

PR #359, opened 2022-10-03
Experiment blueprintjs table

emilyvandewalle:
Experiment to use blueprintjs table in query editor:

image


patrickdevivo:
I'm going to close this PR out for now, as there's been quite a bit of divergence. I will create an issue to track using the blueprintjs table

PR #375, opened 2022-10-07
feat: adding sync type groups

### LINKED ISSUE #379: Add Sync Type Grouping
To address concurrency limits we have with sync we want to introduce a grouping that will allow us to set a concurrency limit for the syncs in that group. The main use case we have for this now is the GitHub API has both a primary and a secondary rate limit which prevent us from making more than one call per second so we are not able to fan out on these syncs. We are however able to do that in other syncs and we would like to allow them to. We currently use a mutex in code to control this but this prevents us from scaling the containers since the mutex can't be shared.

amenowanna:
This PR resolves #379 .

- Introduces a new table called `repo_sync_types_grouping` that will allow us to add groups and concurrency limits for these groups
- The enqueue logic has been updated to allow the window function to break up our enqueue by group so that we don't have to wait for all syncs to finish we just have to wait for syncs in that group to finish. This will allow syncs that can run concurrently to do so and can potentially run several times while syncs in another group are still waiting to be processed
- The dequeue logic now respects the concurrency that is set in `repo_sync_types_grouping` so like in the case of GitHub we only allow one sync in the group at a time so we no longer have to rely on the mutex for this concurrency management
- There is some white space cleanup I did as well


internal/db/queries.sql
```diff
@@ -29,21 +29,33 @@ DO UPDATE SET tags = (
...
+        WHERE status = 'RUNNING'
+),
+dequeued AS (
+   UPDATE mergestat.repo_sync_queue SET status = 'RUNNING'
+   WHERE id IN (   
+        SELECT rsq.id
+        FROM mergestat.repo_sync_queue rsq
+        INNER JOIN mergestat.repo_sync_type_groups rstg ON rsq.type_group = rstg.group
+        WHERE status = 'QUEUED'
+        AND rstg.concurrent_syncs > (SELECT COUNT(*) from running where running.group = rstg.group)
```
patrickdevivo, line 48:
```suggestion
        AND rstg.concurrent_syncs > (SELECT COUNT(*) FROM running WHERE running.group = rstg.group)
```

patrickdevivo (changes requested):

patrickdevivo (approved):

Issue #368, opened 2022-10-05
Create New public tables for GITHUB_ACTIONS entities


ramiroJCB:
relates to #154

Issue #367, opened 2022-10-05
Sync rate limits from http client with our current mergestat-light implementation


ramiroJCB:
relates to #154

PR #395, opened 2022-10-12
feat(actions): First iteration of GITHUB_ACTIONS sync

### LINKED ISSUE #366: Create Http Iterators for GITHUB_ACTIONS Sync
relates to #154

ramiroJCB:
With this pull request we open our first iteration over the new sync type  GITHUB_ACTIONS  this one is  a complex one because itself divides  in three main categories related in order **workflows-> workflow runs -> workflow jobs -> workflow logs** 
Also we  don't use the  graphql client neither  the mergestat-light dependency for this sync .Mainly because we are moving away from mergestat-light dependency  and the graphql api from Github does not provided the information we require for this sync .

To make use of the debug logs in your env file put "DEBUG=1" and re-build your image ,the debug logs  have sufficient information to replicate any of the  http calls that are use in our iterator to test that information externally.

**NOTE* * There are a lot of transient errors from GitHub API that we are not handling in this PR , those errors will be handled in  #408 .

For this specific problem we choosed this algorithm to manage the iteration of getting all github actions 

`
- WorkflowIterator(owner,repo,repoID)
    - Pull a pages of Workflows from the API
        - For each workflows_page
            - HandleWorkflowsUpsert([]workflows ,repoID)
        - WorkflowrunsIterator(owner,repo,[]workflows)
        
- WorkflowrunsIterator(owner,repo,[]workflows,repoID)
    - Pull a page of Workflow runs from the API
        - For each workflowRuns_Page
            - HandleWorkflowRunsUpsert([]workflowRuns,repoID)
        - WorkflowjobsIterator(owner,repo,[]workflowRuns)
        
- WorkflowjobsIterator([]runs)
    - Pull a pages of Workflow jobs from the API
        - For each  []jobs_page
            - Per job we pull a log getWorkflorJobLog(owner,repo,jobID)
            - HandleUpsertWorkflowJob (job, log,repoID)
`
            
  and a picture :) 
resolves 
#366  
#367 
#368

Screen Shot 2022-10-12 at 10 19 26 AM


internal/warehouse/warehouse.go
```diff
@@ -0,0 +1,104 @@
...
+		githubClient: client,
+		logger:       logger,
+		pool:         pool,
+		db:           db,
+	}
+}
+
+func (w *warehouse) restRatelimitHandler(ctx context.Context, resp *github.Response) {
+	var remaining = resp.Rate.Remaining
+	var delay = 0 * time.Millisecond
```
ramiroJCB, line 46: this should be 0.800

internal/warehouse/githubActions.go
```diff
@@ -0,0 +1,383 @@
...
+
+		if err := w.handleUpsertWorkflowJobs(ctx, workflowJob, log, repoID); err != nil {
+			return err
+		}
+
+	}
+	return nil
+}
+
+func (w *warehouse) handleWorflowUpsert(ctx context.Context, workflows []*github.Workflow, repoID uuid.UUID) error {
```
amenowanna, line 209: Name of function has a typo. It is also singular in name but accepts an array. I think based on our conversation the upserts should be singular for all three different upsert functions to keep the same pattern and should manage the looping in the function that calls the upsert.

internal/warehouse/githubActions.go
```diff
@@ -0,0 +1,383 @@
...
+		}); err != nil {
+			return err
+		}
+
+	}
+
+	return tx.Commit(ctx)
+}
+
+func (w *warehouse) handleUpsertWorkflowJobs(ctx context.Context, workflowJob *github.WorkflowJob, log string, repoID uuid.UUID) error {
```
amenowanna, line 322: This is plural but only accepts one job.

migrations/20221012185940_add_repo_sync_type_groups.up.sql
```diff
@@ -0,0 +1,95 @@
+BEGIN;
```
amenowanna, line 1: My comments PR is going to merge first so IMO we should rename this migration so it runs after that one. Agree @patrickdevivo ?

internal/warehouse/githubActions.go
ramiroJCB: Done

internal/warehouse/githubActions.go
ramiroJCB: Done :)

internal/warehouse/githubActions.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+	return nil
+}
+
+func (w *warehouse) handleWorkflowRuns(ctx context.Context, owner, repo string, repoID uuid.UUID, workflowsPage []*github.Workflow) error {
+	var err error
+	var resp *github.Response
+	var workflowRunsPage *github.WorkflowRuns
+	opt := &github.ListWorkflowRunsOptions{
+		ListOptions: github.ListOptions{PerPage: 30},
```
patrickdevivo, line 91: same for this `PerPage` - maybe we address that at a later point

internal/warehouse/githubActions.go
```diff
@@ -0,0 +1,394 @@
...
+			log, err = w.parseJobLogs(workflowJobLog, strconv.Itoa(int(*workflowJob.ID)))
+			if err != nil {
+				return err
+			}
+		}
+
+		if err := os.RemoveAll(strconv.Itoa(i)); err != nil {
+			w.logger.Err(err)
+			return err
+		}
```
patrickdevivo, line 194: So I think the removal of the file should be handled closer to where we create/download the file. In `parseJobLogs`, if we succeed in downloading the file, but then encounter an error reading it (when we call `ReadFile`), then we won't reach this part of the code to cleanup/remove that file, and it will get left behind on disk.

A pattern that could be used is returning a "cleanup" `func` from the `parseJobLogs` which can be called in a `defer` (so that it's always called when the function exits, either with an error or not). This approach is taken [here](https://github.com/mergestat/mergestat/blob/main/internal/syncer/syncer.go#L215) when we clone repos and then clean them up after a sync run (making sure to clean it up even if an error occurs)

My other request here would be that we use a temporary directory to download the file to, again configured similarly to how we do git clones: https://github.com/mergestat/mergestat/blob/main/internal/syncer/syncer.go#L216

as that will allow us to control where these temporary files go on disk (which should be configurable in our deployment somehow)

internal/warehouse/warehouse.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	// Allow for shutdown during the delay
+	select {
+	case <-ctx.Done():
+	case <-time.After(delay):
+	}
+
+}
+func (w *warehouse) getRepoOwnerAndRepoName(repoUrl string) (string, string, error) {
```
patrickdevivo, line 68: do you mind adding a func comment to this just to be extra clear on the behavior and what each string in the output maps to? I assume it's `owner` and `repoName`.

this could make sense as a more general helper too - and one we could write some unit tests against

internal/warehouse/warehouse.go
```diff
@@ -0,0 +1,104 @@
...
+
+	sr = strings.Replace(repoUrl, "https://github.com/", "", -1)
+	s = strings.Split(sr, "/")
+
+	return s[0], s[1], nil
+}
+
+func (w *warehouse) parseJobLogs(logsUrl *url.URL, filepath string) (string, error) {
+	var bytes []byte
+	_, err := grab.Get(filepath, logsUrl.String())
```
patrickdevivo, line 93: I think ideally I'd prefer not to use an external library for download these files. It should be fairly straight-foward using the standard lib: https://stackoverflow.com/questions/11692860/how-can-i-efficiently-download-a-large-file-using-go

It seems like `grab` is more intended for dealing with many concurrent downloads at once, based on the description in the README. In general I think our use of it is narrow here and we should prefer just using our own code to handle

internal/warehouse/warehouse.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		log.Fatal(err)
```
patrickdevivo, line 95:
```suggestion
```

we probably shouldn't panic here

internal/warehouse/warehouse.go
```diff
@@ -0,0 +1,104 @@
...
+		return "", err
+	}
+
+	if bytes, err = ioutil.ReadFile(filepath); err != nil {
+		log.Fatal("Failed to read file: " + filepath)
```
patrickdevivo, line 100:
```suggestion
		return "", fmt.Errorf("failed to read file: %s", filepath)
```

same here

patrickdevivo (changes requested):

internal/warehouse/githubActions.go
```diff
@@ -0,0 +1,394 @@
...
+	w.logger.Info().Msgf("finished getting all github actions")
+	return nil
+}
+
+func (w *warehouse) handleWorkflows(ctx context.Context, owner, repo string, repoID uuid.UUID) error {
+	var err error
+	var resp *github.Response
+	var workflowsPage *github.Workflows
+	opt := &github.ListWorkflowRunsOptions{
+		ListOptions: github.ListOptions{PerPage: 30},
```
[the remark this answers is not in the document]
ramiroJCB: like an env variable ? I could see that happen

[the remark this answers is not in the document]
ramiroJCB: I believe  the env. approach is what you meant

internal/warehouse/githubActions.go
ramiroJCB: For  this iteration we will keep the grap library due to small  issues of our current approach vs the standard  library approach , however we are using the temp pattern now

internal/warehouse/warehouse.go
ramiroJCB: Done

docker-compose.yaml
```diff
@@ -100,11 +103,9 @@ services:
       - ./scripts/grafana/provisioning:/etc/grafana/provisioning
       - ./examples:/var/lib/examples
       - /var/lib/grafana/dashboards/
-    entrypoint: ["bash"]
+    entrypoint: [ "bash" ]
     command: > # This is to set the datasource id in the examples to the datasource created for local grafana
-      -cx 'cp -r /var/lib/examples/* /var/lib/grafana/dashboards/
-      && find /var/lib/grafana/dashboards/ -name ''*.json'' | xargs grep -rnl "$${DS_POSTGRES}" | xargs -i@ sed -i "s/\$${DS_POSTGRES}/SXQZgpP7z/g" @
-      && exec grafana-server --config="/etc/grafana/grafana.ini"'
+      -cx 'cp -r /var/lib/examples/* /var/lib/grafana/dashboards/ && find /var/lib/grafana/dashboards/ -name ''*.json'' | xargs grep -rnl "$${DS_POSTGRES}" | xargs -i@ sed -i "s/\$${DS_POSTGRES}/SXQZgpP7z/g" @ && exec grafana-server --config="/etc/grafana/grafana.ini"'
```
patrickdevivo, line 108: @ramiroJCB @amenowanna just wanna make sure the changes here are intentional - not super familiar with this part of the `docker-compose`

internal/db/queries.sql
```diff
@@ -156,3 +156,230 @@ SELECT @repoID::uuid, type, priority, true
...
+		workflow_url=EXCLUDED.workflow_url,
+		repository_url=EXCLUDED.repository_url,
+		head_repository_url=EXCLUDED.head_repository_url
+  RETURNING xmax::text
+)
+SELECT
+	COUNT(*) AS all_rows,
+    SUM(CASE WHEN xmax::int = 0 THEN 1 ELSE 0 END) AS ins,
+    SUM(CASE WHEN xmax::int > 0 THEN 1 ELSE 0 END) AS upd
+FROM t;
```
patrickdevivo, line 307:
```suggestion
SELECT
    COUNT(*) AS all_rows,
    SUM(CASE WHEN xmax::int = 0 THEN 1 ELSE 0 END) AS ins,
    SUM(CASE WHEN xmax::int > 0 THEN 1 ELSE 0 END) AS upd
FROM t;
```

internal/warehouse/githubActions.go
```diff
@@ -0,0 +1,418 @@
...
+	"net/url"
+
+	"github.com/google/go-github/v47/github"
+	"github.com/google/uuid"
+	"github.com/jackc/pgtype"
+	"github.com/jackc/pgx/v4"
+	"github.com/mergestat/mergestat/internal/db"
+)
+
+func (w *warehouse) GithubActions(ctx context.Context, j *db.DequeueSyncJobRow) error {
```
patrickdevivo, line 16:
```suggestion
func (w *warehouse) GitHubActions(ctx context.Context, j *db.DequeueSyncJobRow) error {
```

a small nit - generally we try to capitalize the `H` in GitHub as that's how they ask to be spelled

internal/warehouse/githubActions.go
```diff
@@ -0,0 +1,418 @@
...
+	w.logger.Info().Msgf("finished getting all github actions")
+	return nil
+}
+
+func (w *warehouse) handleWorkflows(ctx context.Context, owner, repo string, repoID uuid.UUID) error {
+	var err error
+	var resp *github.Response
+	var workflowsPage *github.Workflows
+	pagination, err := w.getPaginationOpt("GITHUB_WORKFLOW_PER_PAGE")
+
```
patrickdevivo, line 51:
```suggestion
```

internal/warehouse/githubActions.go
```diff
@@ -0,0 +1,418 @@
...
+	}
+	return nil
+}
+
+func (w *warehouse) handleWorkflowRuns(ctx context.Context, owner, repo string, repoID uuid.UUID, workflowsPage []*github.Workflow) error {
+	var err error
+	var resp *github.Response
+	var workflowRunsPage *github.WorkflowRuns
+	pagination, err := w.getPaginationOpt("GITHUB_WORKFLOW_RUNS_PER_PAGE")
+
```
patrickdevivo, line 98:
```suggestion
```

internal/warehouse/githubActions.go
```diff
@@ -0,0 +1,418 @@
...
+}
+
+func (w *warehouse) handleWorkflowJobLogs(ctx context.Context, owner, repo string, repoID uuid.UUID, workflowRunJobsPage []*github.WorkflowJob) error {
+	var err error
+	var resp *github.Response
+	var log string
+	var workflowJobLog *url.URL
+	// we create a  tmp dir to store all downloaded files into it
+	filepath, cleanup, err := w.createTempDirForGitClone()
+
```
patrickdevivo, line 209:
```

File: internal/warehouse/warehouse.go [hunk 18]
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	if bytes, err = ioutil.ReadFile(filepath); err != nil {
+		return "", fmt.Errorf("failed to read file: %s", filepath)
+	}
+
+	return string(bytes), nil
+}
+
+func (w *warehouse) createTempDirForGitClone() (string, func(), error) {
```
patrickdevivo, line 115:
```suggestion
func (w *warehouse) createTempDirForJobLogs() (string, func(), error) {
```

renaming this because the purpose is no longer for git clones, but job log file downloads

internal/warehouse/warehouse.go
```diff
@@ -0,0 +1,219 @@
...
+	tmpPath, err := os.MkdirTemp(os.Getenv("GIT_WORKFLOW_LOGS_PATH"), "")
+
```
patrickdevivo, line 117:
```suggestion
```

internal/warehouse/warehouse.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	return tmpPath, func() {
+		if err := os.RemoveAll(tmpPath); err != nil {
+			w.logger.Err(err).Msgf("error cleaning up repo at: %s, %v", tmpPath, err)
+		}
+	}, nil
+}
+
+func (w *warehouse) getPaginationOpt(pagination string) (int, error) {
```
patrickdevivo, line 129: mind adding a quick function comment here to describe what this does?

internal/warehouse/warehouse.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	return pgtype.JSONB{
+		Bytes:  bytes,
+		Status: pgtype.Present,
+	}, nil
+
+}
+
+func (w *warehouse) stringToSqlnullString(v *string) sql.NullString {
```
patrickdevivo, line 167:
```suggestion
func (w *warehouse) stringToSqlNullString(v *string) sql.NullString {
```

(to be consistent with the other capitalizations)

patrickdevivo (changes requested):

internal/db/queries.sql
```diff
@@ -156,3 +156,230 @@ SELECT @repoID::uuid, type, priority, true
...
+	repository_url,
+	head_repository_url)
+	VALUES(
+ 	@repo_id::UUID,
+	@id,
+	@workflowRunNodeId,
+    @name,
+	@headBranch,
+	@runNumber,
+	@runAttemp,
```
patrickdevivo, line 247: @ramiroJCB i think this is a typo - should be `@runAttempt`? (missing `t` at the end)

patrickdevivo (changes requested):

docker-compose.yaml
ramiroJCB: I'm pretty sure than this is a forma issue .re-formatting it again

internal/db/queries.sql
ramiroJCB: done

internal/warehouse/warehouse.go
ramiroJCB: done

internal/warehouse/warehouse.go
ramiroJCB: done

internal/warehouse/warehouse.go
ramiroJCB: done

internal/warehouse/githubActions.go
ramiroJCB: done

internal/warehouse/githubActions.go
ramiroJCB: done

internal/warehouse/githubActions.go
ramiroJCB: done

internal/db/queries.sql
patrickdevivo: this one seems to still be here @ramiroJCB

internal/warehouse/githubActions.go
```diff
@@ -0,0 +1,416 @@
+package warehouse
+
```
patrickdevivo, line 2: can we rename this file `github_actions.go` to be consistent with go file naming (and how we do it elsewhere in the repo)

patrickdevivo (changes requested):

internal/db/queries.sql
ramiroJCB: now should be done

internal/warehouse/githubActions.go
ramiroJCB: done

patrickdevivo (approved):

PR #414, opened 2022-10-20
feat: add metrics to repo overview

### LINKED ISSUE #222: Consider some "top level" stats/metrics to display in `/repos` page?
Above the main repos table, maybe some things like:

- Total number of repos (that match the current filters - tags and search terms)
- Avg. number of syncs per repo
- Avg. sync run time

etc. Basically some numbers (maybe small charts?) to give the user an aggregate sense of what their repo and sync data looks like

emilyvandewalle:
- Started adding top metrics
- I was able to figure out the `Number of repos` and `Avg number of syncs`, but not `Avg sync time` for which I could use some help

image

(Resolves #222)

### DIFF
--- a/ui/src/views/repositories/index.tsx
+++ b/ui/src/views/repositories/index.tsx
@@ -19,7 +19,7 @@ interface MetricNumberProp {
 
 const MetricNumber: React.FC<MetricNumberProp> = ({ loading, metric }: MetricNumberProp) => {
   return (
-    <div className='h-10'>
+    <div>
       {loading ? <Spinner size="sm" /> : metric}
     </div>
   )
@@ -87,8 +87,8 @@ const RepositoriesView: React.FC = () => {
               <Stat.Left>
                 <Stat.Label>Latest syncs with errors</Stat.Label>
                 <Stat.Number>
-                  <div className='flex space-x-1.5'>
-                    {metrics && metrics?.totalRepoSyncsError > 0 && <CircleErrorFilledIcon className='t-icon t-icon-danger mt-1' />}
+                  <div className='flex space-x-1.5 items-center'>
+                    {<CircleErrorFilledIcon className='t-icon t-icon-danger' />}
                     <MetricNumber loading={loading} metric={metrics?.totalRepoSyncsError || 0} />
                   </div>
                 </Stat.Number>
--- a/ui/src/views/repositories/index.tsx
+++ b/ui/src/views/repositories/index.tsx
@@ -51,7 +51,7 @@ const RepositoriesView: React.FC = () => {
             <Stat className='shadow-sm w-full'>
               <Stat.Left>
                 <Stat.Label>Total repos</Stat.Label>
-                <Stat.Number>{data?.repos?.totalCount}</Stat.Number>
+                <Stat.Number>{data?.allRepos?.totalCount}</Stat.Number>
               </Stat.Left>
               <Stat.Right>
                 <ColoredBox size='12'><RepositoryIcon className='t-icon t-icon-default' /></ColoredBox>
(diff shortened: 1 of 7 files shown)


patrickdevivo:
This looks great! @amenowanna and I discussed a bit and we wanted to change the numbers a bit. Implementation may need to involve @german-mergestat too, so tagging him in. But overall look and feel wise looks great.

The numbers should be:
- Total number of repos (stays the same)
- Total number of repo syncs (i.e. the sum of all the "sum" values in the table)
- Total number of repo syncs where the latest run had an error (i.e. the sum of all the "error" values in the table)

The other ask is to see if we can move the repo search bar to _below_ these numbers, to make it clear that they are global totals (and that searching does not change them). Thanks so much @emilyvandewalle

amenowanna:
Looks great! My only comment here is that the titles of each metric is using different wording. Can we make it consistent please. Like `Total number of...` or `Number of...` @patrickdevivo have a preference?

emilyvandewalle:
Oops, good catch, agree this should be consistent, adjusted it to `Total number of...` ! 👍

emilyvandewalle:
(I was a bit too fast in trying to adjust it, I'll wait for @patrickdevivo to give his preference)

patrickdevivo:
(Copied from Slack)

- Let’s change the text to just say Total repos Total repo syncs and Latest syncs with errors
- Let’s add a loading/indeterminate state for  the “metric box” when the data is fetching from the API (you can trigger this by writing a search query - tho these should be unrelated, I think we should actually pull them out into separate queries, especially since pagination will change that list-repos query)
- Finally the last box (Latest syncs with errors) needs to be implemented data wise

I think once those are addressed it should be good to merge!

patrickdevivo:
```graphql
query {
  repos {
    totalCount
  }
  repoSyncs(condition: {scheduleEnabled: true}) {
    totalCount
  }
  execSQL(query: "WITH ranked_completed_syncs AS ( SELECT rsq.id, rsq.repo_sync_id, rsq.done_at, RANK() OVER(PARTITION BY rsq.repo_sync_id ORDER BY rsq.done_at DESC) AS rank_num FROM mergestat.repo_sync_queue AS rsq WHERE rsq.repo_sync_id NOT IN (SELECT repo_sync_id FROM mergestat.repo_sync_queue WHERE status = 'RUNNING' OR status = 'QUEUED') ) SELECT COUNT(DISTINCT rcs.id) AS syncs_error_count FROM ranked_completed_syncs AS rcs INNER JOIN mergestat.repo_sync_logs AS rsl ON rcs.id = rsl.repo_sync_queue_id WHERE rank_num = 1 AND log_type = 'ERROR'")
}
```

patrickdevivo:
image

ui/src/views/repositories/index.tsx
```diff
@@ -2,43 +2,113 @@ import { useRepositoriesContext } from 'src/state/contexts/repositories.context'
...
+                  <MetricNumber loading={loading} metric={metrics?.totalRepoSyncs || 0} />
+                </Stat.Number>
+              </Stat.Left>
+              <Stat.Right>
+                <ColoredBox size='12'><TableIcon className='t-icon t-icon-default' /></ColoredBox>
+              </Stat.Right>
+            </Stat>
+            <Stat className='shadow-sm w-full'>
+              <Stat.Left>
+                <Stat.Label>Latest syncs with errors</Stat.Label>
```
patrickdevivo, line 88:
```suggestion
                <Stat.Label>Total syncs with errors</Stat.Label>
```

patrickdevivo (changes requested):

ui/src/views/repositories/index.tsx
german-mergestat: Done [4d2715a](https://github.com/mergestat/mergestat/pull/414/commits/4d2715a2d33d47e1f4d444176bcbe34426c762e9)

patrickdevivo (approved):

PR #416, opened 2022-10-20
feat: test moving alert stack to bottom

emilyvandewalle:
image


patrickdevivo:
Going to close this out, as it was addressed recently in `blocks`

PR #419, opened 2022-10-21
fix: remove `.env-worker` requirement for `docker-compose`

### LINKED ISSUE #418: Missing .env-worker file
When trying to run the `docker-compose up` I get the error:
open ../.env-worker: no such file or directory

It seems like this file is missing in the repository, and there is no instruction on what should it contain.

patrickdevivo:
For now, just use inline env vars in the `docker-compose.yaml` file. We can configure a `make` target for a more complex setup that uses `.env-worker` later if we need it.

This allows the app to be brought up on a "fresh" clone of the repo with just `docker-compose up`. Closes #418


amenowanna (approved):

asheone:
Lovely stuff. Thanks for the quick resolution!

PR #429, opened 2022-10-24
fix: set readaccess defaults 

### LINKED ISSUE #428: Update postgres permissions to use defaults
We currently do not use postgres defaults which allows permissions to be granted for any new object that is created by another user for the user/role we want to grand privileges too. Now for example the `readaccess` role does not have select permissions to tables that were created after this role was granted privileges like `git_blame`.

amenowanna:
Added defaults for readaccess select privileges. Changed the migrations numbers to be a sequence instead of date driven to address bad name given on previous migration as well as to force us to use a sequence and enforce merging of migrations when there is a conflict.

Tested by creating a new table in public schema and being able to still select it from the Queries section in the UI.
image

resolves #428

### DIFF
--- a/migrations/900000000000001_readaccess_defaults.up.sql
+++ b/migrations/900000000000001_readaccess_defaults.up.sql
@@ -1,0 +1,26 @@
+-- https://stackoverflow.com/questions/8092086/create-postgresql-role-user-if-it-doesnt-exist
+-- idempotent role creation
+DO
+$do$
+BEGIN
+IF EXISTS (
+  SELECT FROM pg_catalog.pg_roles
+  WHERE  rolname = 'readaccess') THEN
+  RAISE NOTICE 'Role "readaccess" already exists. Skipping.';
+ELSE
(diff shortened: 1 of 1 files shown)


migrations/900000000000001_readaccess_defaults.up.sql
```diff
@@ -0,0 +1,26 @@
...
+BEGIN
+IF EXISTS (
+  SELECT FROM pg_catalog.pg_roles
+  WHERE  rolname = 'readaccess') THEN
+  RAISE NOTICE 'Role "readaccess" already exists. Skipping.';
+ELSE
+  CREATE ROLE readaccess;
+END IF;
+END
+$do$;
```
patrickdevivo, line 14: any reason this repeats https://github.com/mergestat/mergestat/blob/main/migrations/20220811202258_pg_read_only_user.up.sql ?

patrickdevivo (dismissed):

patrickdevivo (approved):

PR #484, opened 2022-10-31
feat(helper):Adding helper fns to a helper package

### LINKED ISSUE #462: We should move all our helper functions to helper package
There are a lot of  untested helper functions across the worker that should move to a single package and be tested  in a unit test manner .With this approach we can say that anything that is inside that package should be tested because functionality should  simple .

ramiroJCB:
Moving all helper functions to one single helper package , not only this a more organize approach but also we are getting our first step to unit testing in the worker :) 
resolves #462


internal/helper/sqlMutations.go
```diff
@@ -0,0 +1,111 @@
+package helper
```
patrickdevivo, line 1: Let's name this file `sql_conversions.go` (underscore for consistency, "conversions" not tied to, but sounds clearer to me this is helpers for type conversions)

patrickdevivo (changes requested):

internal/syncer/git_commit_stats.go
```diff
@@ -68,18 +70,24 @@ type commitStat struct {
...
-	tmpPath, cleanup, err := w.createTempDirForGitClone(j)
+	tmpPath, cleanup, err := helper.CreateTempDir(os.Getenv("GIT_CLONE_PATH"), "mergestat-repo-")
 	if err != nil {
 		return fmt.Errorf("temp dir: %w", err)
 	}
-	defer cleanup()
+	defer func() {
+		if err = cleanup(); err != nil {
+			l.Err(err).Msgf("error cleaning up repo at: %s, %v", tmpPath, err)
+
```
patrickdevivo, line 88:
```suggestion
```

patrickdevivo (approved):

PR #488, opened 2022-11-01
feat: clean up `docker-compose` a bit

patrickdevivo:
and add `make` targets for easier building and a note to `docs/README.md`


ramiroJCB:
Are we adding the docker-compose down in this pr or in another ?

ramiroJCB (approved):

PR #492, opened 2022-11-02
chore(metrics): Splitting metric queries

### LINKED ISSUE #485: Split up GraphQL requests on `/repos` page
The `/repos` page makes 4 (I think) separate GraphQL queries all in one request - 3 for the "metric" numbers and another to retrieve the table of repos.

Image

Let's split these up into 4 separate API requests, in case one takes long to run (or has an error/times out), that way the _entire_ page doesn't fail to load, only the section with that query.

german-mergestat:
Splitting metric queries. (Resolves #485)

metrics

### DIFF
--- a/ui/src/api-logic/graphql/queries/get-repos.query.ts
+++ b/ui/src/api-logic/graphql/queries/get-repos.query.ts
@@ -1,5 +1,27 @@
 import { gql } from '@apollo/client'
 
+const GET_ALL_REPOS = gql`
+  query getAllRepos {
+    allRepos: repos {
+      totalCount
+    }
+  }
+`
+
+const GET_ALL_ENABLED_REPOS = gql`
+  query getAllEnabledRepos {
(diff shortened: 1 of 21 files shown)


ui/src/api-logic/graphql/queries/get-repos.query.ts
```diff
@@ -1,20 +1,35 @@
 import { gql } from '@apollo/client'
 
-const GET_REPOS = gql`
-  query getRepos($search: String!, $first: Int, $offset: Int) {
-    serviceAuthCredentials(filter: {type: {equalTo: "GITHUB_PAT"}}) {
+const GET_ALL_REPOS = gql`
```
patrickdevivo, line 3:
```suggestion
const GET_REPO_TOTAL_COUNT = gql`
```

Just to be a bit clearer

patrickdevivo (changes requested):

patrickdevivo (approved):

PR #509, opened 2022-11-04
feat(logs): Implementing copy to clipboard for logs

### LINKED ISSUE #479: Copy button in logs does not work
If the fix is simple, let's implement it - otherwise, let's just remove/hide the button for now

image

german-mergestat:
Implementing copy to clipboard for logs. (Resolves #479)

copy-logs

### DIFF
--- a/ui/src/utils/index.tsx
+++ b/ui/src/utils/index.tsx
@@ -141,6 +141,14 @@ export const copy = (text: string | null | undefined) => {
 }
 
 /**
+ * Method to copy to clipboard a given strings array
+ * @param array Strings array to process
+ */
+export const copyArrayToClipboard = (array: string[] = []) => {
+  copy(array.join('\n'))
+}
+
+/**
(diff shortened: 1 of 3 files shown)


ui/src/utils/index.tsx
```diff
@@ -140,6 +140,14 @@ export const copy = (text: string | null | undefined) => {
   showSuccessAlert('Copied')
 }
 
+/**
+ * Method to copy to clipboard a given strings array
+ * @param array Strings array to process
+ */
+export const copyArrayToClipboard = (array: string[] = []) => {
+  copy(array.join('\n'))
```
patrickdevivo, line 148: Is this somewhere it could make sense to use the PapaParse library?

patrickdevivo (approved):

ui/src/utils/index.tsx
german-mergestat: No so sure, I just was trying use `unparse` but seems it does not work for array of strings.

PR #515, opened 2022-11-08
feat: last error query performance enhancements

### LINKED ISSUE #483: `/repos` page timing out on deployments with many repos (and big sync history)
We received a report from a user with 800+ repos imported and many active syncs that the console times out when loading `/repos`. My _guess_ without knowing much more is that our query to display the `Total syncs with errors` number is taking a long time to run (or at least that's the first place I'd look).

image

We might need to optimize somewhere to ensure whatever query is causing the timeout/slow response is speedy.

amenowanna:
Changing logic to store the queue id for the last completion of each repo sync. This will allow us to check for errors with in the sync runs without having to rank all of the jobs in the queue which is a very expensive query. I tested with several different indexing changes but our current indexing continued to be chosen by the database engine. The most expensive part of this query is still having to check the queue to not query syncs that are currently `QUEUED` or `RUNNING`.

Resolves #483

### DIFF
--- a/migrations/900000000000003_add_set_sync_job_status.up.sql
+++ b/migrations/900000000000003_add_set_sync_job_status.up.sql
@@ -33,10 +33,10 @@ RETURNS UUID
 AS
 $$
 DECLARE _repo_sync_id UUID;
-begin
+BEGIN
     IF new_status = 'DONE' THEN
             WITH update_queue AS (
-                UPDATE mergestat.repo_sync_queue SET "status" = new_status where mergestat.repo_sync_queue.id = repo_sync_queue_id
+                UPDATE mergestat.repo_sync_queue SET "status" = new_status WHERE mergestat.repo_sync_queue.id = repo_sync_queue_id
                 RETURNING *
             )
             UPDATE mergestat.repo_syncs set last_completed_repo_sync_queue_id = repo_sync_queue_id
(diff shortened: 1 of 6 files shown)


migrations/900000000000003_add_set_sync_job_status.up.sql
```diff
@@ -0,0 +1,55 @@
...
+
+CREATE OR REPLACE FUNCTION mergestat.set_sync_job_status(new_status TEXT, repo_sync_queue_id BIGINT)
+RETURNS UUID
+AS
+$$
+DECLARE _repo_sync_id UUID;
+begin
+    IF new_status = 'DONE' THEN
+            WITH update_queue AS (
+                UPDATE mergestat.repo_sync_queue SET "status" = new_status where mergestat.repo_sync_queue.id = repo_sync_queue_id
```
patrickdevivo, line 39:
```suggestion
                UPDATE mergestat.repo_sync_queue SET "status" = new_status WHERE mergestat.repo_sync_queue.id = repo_sync_queue_id
```

migrations/900000000000003_add_set_sync_job_status.up.sql
```diff
@@ -0,0 +1,55 @@
...
+FROM ranked_completed_syncs AS rcs
+WHERE rcs.rank_num = 1
+    AND rcs.repo_sync_id = mergestat.repo_syncs.id;
+
+CREATE OR REPLACE FUNCTION mergestat.set_sync_job_status(new_status TEXT, repo_sync_queue_id BIGINT)
+RETURNS UUID
+AS
+$$
+DECLARE _repo_sync_id UUID;
+begin
```
patrickdevivo, line 36:
```suggestion
BEGIN
```

migrations/900000000000003_add_set_sync_job_status.up.sql
```diff
@@ -0,0 +1,55 @@
...
+            FROM update_queue
+            WHERE mergestat.repo_syncs.id = update_queue.repo_sync_id
+            RETURNING mergestat.repo_syncs.id INTO _repo_sync_id;
+    ELSE    
+            UPDATE mergestat.repo_sync_queue SET "status" = new_status where mergestat.repo_sync_queue.id = repo_sync_queue_id
+            RETURNING repo_sync_id INTO _repo_sync_id;
+    END IF;
+    
+    RETURN _repo_sync_id;    
+end;
```
patrickdevivo, line 52:
```suggestion
END;
```

patrickdevivo (changes requested):

patrickdevivo (approved):

PR #518, opened 2022-11-08
chore(loaders): Adding feedback actions for the user

### LINKED ISSUE #473: Trigger user-feedback (loaders) when actions are taken in UI
In some areas of the console we don't give feedback to the user that an action was taken, and if the API request is not immediate, it feels like no action was taken. Similarly, if we need to re-fetch data after something is _created_ (so that it shows up), we should implement that (with a loading state to show that we're refetching).

- [ ] Adding an auto import (it takes a moment for feedback in the UI - waits for the poll interval on the `/repos` page to refetch)
- [ ] Sync now buttons (there are 2 sync now buttons, clicking them should trigger a loading state + the mutation)
- [ ] The auto imports in the settings - loading the page should trigger a refetch (we visited the page, went back to `/repos`, created another repo auto import, went _back_ to the "repo auto import settings" area, and the new import didn't show up (need to trigger a refetch)

german-mergestat:
Adding feedback actions for the user. (Resolves #473)

Fixed:
- Adding fast feedback when sync button is pressed (leaving it in queued state)
- Showing an alert when schedule toggle is switched. (Sync On/Sync Off)

Couldn't reproduce them:
- Blue banner with the spinner is being showed immediately.
- Auto import list in settings is being showed well if a new import is added. For this one I remember I did a fix.

### DIFF
--- a/ui/src/views/repository-data/components/sync-types-table/index.tsx
+++ b/ui/src/views/repository-data/components/sync-types-table/index.tsx
@@ -78,7 +78,7 @@ export const SycnTypesTable: React.FC<SycnTypesTableProps> = ({ repoId, data }: 
                           <Toggle
                             isChecked={sync.data.scheduleEnabled}
                             onChange={() => {
-                              showSuccessAlert(`Sync ${sync.data.scheduleEnabled ? 'Off' : 'On'}`)
+                              showSuccessAlert(`Sync schedule turned ${sync.data.scheduleEnabled ? 'off' : 'on'}`)
                               sync.data.id
                                 ? updateSchedule({ variables: { syncId: sync.data.id, schedule: !sync.data.scheduleEnabled } })
                                 : addSyncType({ variables: { repoId, syncType: sync.data.type, schedule: true } })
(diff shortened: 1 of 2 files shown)


ui/src/views/repository-data/components/sync-types-table/index.tsx
```diff
@@ -77,6 +78,7 @@ export const SycnTypesTable: React.FC<SycnTypesTableProps> = ({ repoId, data }:
                           <Toggle
                             isChecked={sync.data.scheduleEnabled}
                             onChange={() => {
+                              showSuccessAlert(`Sync ${sync.data.scheduleEnabled ? 'Off' : 'On'}`)
```
patrickdevivo, line 81:
```suggestion
                              showSuccessAlert(`Sync schedule turned ${sync.data.scheduleEnabled ? 'off' : 'on'}`)
```

patrickdevivo (changes requested):

patrickdevivo (approved):

PR #543, opened 2022-11-17
feat(worker):Implementing testing pattern in github actions 

### LINKED ISSUE #539: Adding unit test to github_actions sync
First implementation of  our testing pattern.To complete this item we need to :

- [x]  use our pattern to create mocks for pgx pool,db queries and  pgx transaction type
- [x]  Update our sqlc.yaml file to get interfaces from our queries
- [x]  Change the way we pass down our db and pool into the worker
- [x]  Test for nil pointer exceptions in all three  entities  of the actions(workflows,runs and jobs)

ramiroJCB:
Here we implement our first  iteration for our test pattern
To test this piece  of work  you can go for one of the two routes:

1. CLI , go and check if you have golang 1.19. if not get it , then run `go test ./....` and all test should run :) 
2. VScode in your activity bar check for the testing option , then  go for either github/mergestat/mergestat or github_actions.go folders and run the play button and all test should run 

To verify that the tests are running againts nil validation , please go to mergestat/internal/helper/helper.go
and comment lines 57 to 60 .Now  test  that do this convertions from int to int32 should fail because of a nil pointer exception 

resolves #539


amenowanna:
@ramiroJCB thank you for addressing all of my comments. I am going to allow @patrickdevivo to be the one to complete the review and approve. It was so great to do actual debugging locally and stepping through code as well as seeing the tests pass and fail. The tests are very fast too! 
image

internal/warehouse/github_actions.go
```diff
@@ -455,3 +456,19 @@ func (w *warehouse) handleWorkflowJobUpsert(ctx context.Context, workflowJob *gi
...
+	return githubClient.Actions.ListWorkflowRunsByID(ctx, owner, repo, workflowID, opt)
+}
+
+var getWorkflowJobs = func(ctx context.Context, githubClient *github.Client, owner, repo string, workflowRunID int64, opt *github.ListWorkflowJobsOptions) (*github.Jobs, *github.Response, error) {
+	return githubClient.Actions.ListWorkflowJobs(ctx, owner, repo, workflowRunID, opt)
+}
+
+var getWorkflowJobLog = func(ctx context.Context, githubClient *github.Client, owner, repo string, workflowJobID int64) (*url.URL, *github.Response, error) {
+	return githubClient.Actions.GetWorkflowJobLogs(ctx, owner, repo, workflowJobID, true)
+}
```
patrickdevivo, line 474: Is there a reason to define these as variables and not just functions?

```go
func getWorkflows(ctx context.Context, githubClient *github.Client, owner, repo string, opt *github.ListWorkflowRunsOptions) (*github.Workflows, *github.Response, error) {
  ...
}
```

ramiroJCB: not really , I was at the beguining  trying a pattern related to the var option.But now are not needed

internal/warehouse/github_actions_test.go
```diff
@@ -0,0 +1,443 @@
...
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v4"
+	"github.com/mergestat/mergestat/internal/db"
+	"github.com/mergestat/mergestat/internal/helper"
+	"github.com/mergestat/mergestat/internal/mocks"
+	"github.com/migueleliasweb/go-github-mock/src/mock"
+	"github.com/rs/zerolog"
+)
+
+func Test_warehouse_handleWorkflows(t *testing.T) {
```
patrickdevivo, line 21: Let's follow the Go conventions here on all these test names please: https://go.dev/doc/effective_go#mixed-caps

patrickdevivo (changes requested):

internal/warehouse/github_actions_test.go
ramiroJCB: nice catch , I forgot about this one :)

patrickdevivo (approved):

patrickdevivo (approved):

PR #544, opened 2022-11-17
feat: refactor the `execSQL` implementation

### LINKED ISSUE #423: Support query results with multiple columns with the same name
The query in the screenshot is valid SQL that returns a result set with 2 columns with the same name. However, the GraphQL API returns each row as a JSON object, making it impossible to express a table with multiple columns of the same name (only one is returned).

image

In order to support this, we will likely need to change the format of the response data (possibly an array to hold the column name info and then an array of arrays for the results)

patrickdevivo:
Use `rowMode: 'array'` approach, which closes #423 (see for more context). This approach should allow us to return results with columns that have the same name.

image


amenowanna (dismissed):

amenowanna (approved):
Tested everything locally and it is working great! 🎉

german-mergestat (approved):

PR #552, opened 2022-11-21
feat(import):Managing auto import errors

### LINKED ISSUE #249: Figure out how to convey `repo auto import` failures to the user
Currently, if a repo auto import fails, nothing is conveyed to the user through the UI (because there's nothing in the DB/API to convey this). The only place it's reported is in the worker logs. This is not a great UX, especially because it's quite easy to trigger a failure state (if an org or username is mistyped for instance).

This ticket is for figuring out how to begin tracking repo import errors in the DB (emitted by the worker) so that we can eventually show it in the UI in some way.

ramiroJCB:
here is our first iteration for managing impor errors in the worker .
To test this implementation please do:

1. make docker-clean
2. make dev
3. loging into the mergestat
4. add credentials
5. add an automatic import with a unexisting organization or user
6. check the logs  for warn type msg   to see the error ocurred
7. use the ID given in the logs to search that specific import failure 
8. :)
Screen Shot 2022-11-21 at 2 26 01 PM
Screen Shot 2022-11-21 at 2 26 52 PM
 

resolves #249


internal/db/queries.sql
```diff
@@ -18,6 +18,9 @@ WITH dequeued AS (
 SELECT id, created_at, updated_at, type, settings FROM dequeued;
 ;
 
+-- name: UpdateImportStatus :exec
+UPDATE mergestat.repo_imports SET import_status= @status::TEXT ,import_error = @error::TEXT WHERE id = @ID;
```
patrickdevivo, line 22:
```suggestion
UPDATE mergestat.repo_imports SET import_status= @status::TEXT, import_error = @error::TEXT WHERE id = @ID;
```

ramiroJCB: nice catch

internal/db/queries.sql
```diff
@@ -18,6 +18,9 @@ WITH dequeued AS (
 SELECT id, created_at, updated_at, type, settings FROM dequeued;
 ;
 
+-- name: UpdateImportStatus :exec
+UPDATE mergestat.repo_imports SET import_status= @status::TEXT,import_error = @error::TEXT WHERE id = @ID;
```
patrickdevivo, line 22:
```suggestion
UPDATE mergestat.repo_imports SET import_status = @status::TEXT, import_error = @error::TEXT WHERE id = @ID;
```

patrickdevivo (approved):

PR #555, opened 2022-11-22
fix: use `encodeURIComponent` instead of `encodeURI`

### LINKED ISSUE #453: Validate special characters in passwords when logging in
If a db user has special characters in their password, they should still be able to login. Validate this behavior and apply any fixes we need

patrickdevivo:
in the login API. There are 11 characters (according to this page: https://thisthat.dev/encode-uri-vs-encode-uri-component/) not encoded by `encodeURIComponent` which may appear in PG passwords, causing issues logging in for users with those in their password.

This should address that, closes #453

### DIFF
--- a/ui/src/pages/api/admin-auth.ts
+++ b/ui/src/pages/api/admin-auth.ts
@@ -28,13 +28,12 @@ const adminAuth = async (req: NextApiRequest, res: NextApiResponse) => {
 
     // override the user and password in the connection string with the ones provided by the client
     url.username = user
-    url.password = password
+    url.password = encodeURIComponent(password) // password may contain special characters
 
     try {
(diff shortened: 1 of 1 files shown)


ramiroJCB:
@patrickdevivo  hey how do we test this change ?

patrickdevivo:
You can create a user in the DB and then change the password to include various special characters. You can then validate logging in via the local UI:

```sql
SELECT mergestat.user_mgmt_add_user('user', 'pass', 'ADMIN');
SELECT mergestat.user_mgmt_update_user_password('user', 'pass$')
```

ramiroJCB (approved):

german-mergestat (approved):

PR #556, opened 2022-11-22
feat(users): Implementing user list functionality

german-mergestat:
Implementing user list functionality.

- User list
- Searching input
- Unit Testing

users

### DIFF
--- a/ui/src/api-logic/mappers/users.ts
+++ b/ui/src/api-logic/mappers/users.ts
@@ -15,7 +15,7 @@ const mapToUserData = (data: UsersQuery | undefined): Array<UserData> => {
       // Consolidated User info
       const userInfo: UserData = {
         name: usr.rolname || '',
-        roles: usr.memberof.map(getRole)
+        role: getRole(usr.memberof)
       }
       mappedData.push(userInfo)
     }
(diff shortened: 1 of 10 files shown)


ui/src/api-logic/mappers/users.ts
```diff
@@ -0,0 +1,38 @@
...
+      }
+      mappedData.push(userInfo)
+    }
+  })
+
+  return mappedData
+}
+
+const getRol = (rol: Array<string | null>) => {
+  switch (rol[0]) {
```
patrickdevivo, line 28: @german-mergestat I think we need to check all roles in the `memberof` here - it's possible the user may have other roles and that the ones we're looking for are not the first (or only). We should check for the presence of the role in the whole list.

patrickdevivo (changes requested):

ui/src/api-logic/mappers/users.ts
german-mergestat: Done [9f3646c](https://github.com/mergestat/mergestat/pull/556/commits/9f3646cfb0f051f1cd97afd0b932de6f376f60ef)

ui/src/api-logic/mappers/users.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const getRole = (role: string | null) => {
+  switch (role) {
+    case 'mergestat_role_admin':
+      return USER_TYPE.ADMIN
+    case 'mergestat_role_user':
+      return USER_TYPE.USER
+    default:
+      return USER_TYPE.READ_ONLY
+  }
```
patrickdevivo, line 35: I think we should be careful about defaulting to the `READ_ONLY` role here - since it means something specific. If a role cannot be determined, I think we should display `Unknown` in the UI (and maybe have that as a type here).

patrickdevivo (changes requested):

ui/src/api-logic/mappers/users.ts
german-mergestat: Done [5c1d507](https://github.com/mergestat/mergestat/pull/556/commits/5c1d507283340088fdb99a37e8fdbab9a7320ade)

patrickdevivo (approved):

PR #573, opened 2022-11-26
chore(users): Implementing change password in settings

german-mergestat:
Implementing change password in settings.

- Unit testing

change-pass


ui/src/views/settings/modals/change-password-modal/index.tsx
```diff
@@ -1,15 +1,45 @@
...
+  })
+
   const close = useCallback(() => {
     setShowChangePasswordModal(false)
   }, [setShowChangePasswordModal])
 
+  const handleChangePassword = () => {
+    setError('')
+
+    if ((password && password.length < 10) || (passwordConfirm && passwordConfirm.length < 10)) {
```
patrickdevivo, line 33: let's remove this requirement for now - since we don't enforce it on the backend and we don't have it in the update password area of `User Management` either

patrickdevivo (approved):

PR #643, opened 2022-12-12
feat: implement a `gitleaks` syncer for detecting secrets in a repo

patrickdevivo:
Another POC syncer that brings in this tool: https://gitleaks.io/ as a sync type. Runs a scan on a repo and then stores the JSON output in the DB. Also sets up a view to make the JSON data a bit easier to consume.

image

### DIFF
--- a/migrations/900000000000012_add_gitleaks_sync.up.sql
+++ b/migrations/900000000000012_add_gitleaks_sync.up.sql
@@ -1,0 +1,33 @@
+BEGIN;
+
+INSERT INTO mergestat.repo_sync_types (type, description, short_name) VALUES ('GITLEAKS_REPO_SCAN', 'Executes a gitleaks scan on a git repository', 'Gitleaks Repo Scan') ON CONFLICT DO NOTHING;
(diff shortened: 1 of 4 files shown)


migrations/900000000000012_add_gitleaks_sync.up.sql
```diff
@@ -0,0 +1,33 @@
+BEGIN;
+
+INSERT INTO mergestat.repo_sync_types (type, description, short_name) VALUES ('GITLEAKS_REPO_SCAN', 'Executes a gitleaks scan on a git repository', 'Gitleaks Repo Scan') ON CONFLICT DO NOTHING;
```
amenowanna, line 3: If you don't define priority it will default to 0 and that will compete with things like "Sync Now" which we use for top priority.

patrickdevivo: updated to default to `priority` `3`, like trivy and syft

amenowanna (approved):

Issue #635, opened 2022-12-09
Add support for local file system repos (in the `docker-compose`/local experience)


patrickdevivo:
Add support for local filesystem Git repos in the `docker-compose` MergeStat experience so that local users can point MergeStat at repositories they already have on disk.

We will likely need to produce some instructions on how to mount a host volume on the `worker` service.

riyaz-ali:
This can be achieved by creating an override `docker-compose.yaml` file locally (by the end user) and adding a bind-mount volume.

```yaml
services:
  worker:
    volumes:
      - ~/projects:/projects # this can be any local path
```
Then, from the UI, the repository can be added by using a `file://` url:

![image](https://user-images.githubusercontent.com/17478576/207533264-c17a2319-5ae3-42c3-bb48-157e96b733ea.png)

patrickdevivo:
I had some trouble in this setup when it came to running the `GIT_*` sync types - were you able to get data syncing okay locally @riyaz-ali? I added the `volumes` to the `docker-compose` file locally.

<img width="2048" alt="image" src="https://user-images.githubusercontent.com/57259/207751660-c58c2ea5-6dae-46fa-9001-af1b86f00e95.png">

<img width="2048" alt="image" src="https://user-images.githubusercontent.com/57259/207751814-def70a7b-b734-4fc4-9efa-bb5ccf0f278c.png">

riyaz-ali:
Ah! sorry my bad. You need an additional `/` after the scheme, such that the path reads `/projects/mergestat`. The full url would be `file:///projects/mergestat`

patrickdevivo:
Got it! Thanks - I was able to validate that! Really cool that we can support this in an ok way for now. So I think for this ticket let's mark it complete when we've:

- Documented in our [docs](https://github.com/mergestat/docs) (maybe in the `Running Locally` section? Or `Setup` section? @amenowanna what do you think)
- Documented in the `docker-compose.yaml` file with a comment (let's include an example of how to mount a volume to the worker and a link to the docs page)

Anything else?

amenowanna:
We should put the docs in the running locally section. In general we still need to add docs for adding individual repos and those docs could link to the running local docs for people looking to that.

amenowanna:
@riyaz-ali please finish creating the docs for this so we can get flow moving on this issue again and moved to done.

PR #644, opened 2022-12-14
fix(syncer): do not append token to local URLs

riyaz-ali:
When testing out support for local or `file://` URLs (see #635), I ran into this issue where the GitHub token was being appended (and the path was being escaped / encoded), even for local filesystem paths.


ramiroJCB:
We have a flaky test on the UI , please retry it until it passes

patrickdevivo (approved):

PR #648, opened 2022-12-15
feat(query-editor): Adding toggle for non read only query executions

### LINKED ISSUE #481: Allow for non-read-only queries in the `Queries` UI
Currently, the `Queries` UI only permits read-only queries. This was a limitation put in place to prevent users from taking destructive actions (by accident). However, it turns out that it's quite useful to be able to make "write" statements in the editor, for management actions that's not yet available in the UI.

Consider adding a "non-read-only mode" in the UI (and backend). I'm thinking a toggle with a warning perhaps.

german-mergestat:
Adding toggle for non read only query executions. (Resolves #481)

Read-only enabled:

disabled


amenowanna:
Functionally all my tests worked! I think we should have @patrickdevivo and @emilyvandewalle chime in on the design including the wording on the tooltip. ❤️

patrickdevivo (approved):

PR #650, opened 2022-12-16
feat(sync): Setting typeGroup when syncing

### LINKED ISSUE #633: When selecting "Sync Now" in the console we need to insert type_group to respect concurrency
In the console when we select `Sync Now` we are not setting the type_group value in mergestat.repo_sync_queue. The default for this column is `DEFAULT` and the concurrency for `GITHUB` is then no longer used and we hit the secondary rate limit. We need to add the correct type when we perform this action.

german-mergestat:
Setting `typeGroup` when syncing. (Resolves #633)

### DIFF
--- a/ui/src/@types.tsx
+++ b/ui/src/@types.tsx
@@ -61,6 +61,7 @@ export type RepoSyncDataType = {
     title: string
     brief: string
     scheduleEnabled: boolean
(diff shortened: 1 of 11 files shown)


patrickdevivo (approved):
works as expected locally - thanks!

Issue #578, opened 2022-11-28
Implement "labels" or "topics" for repo sync types


patrickdevivo:
It would be nice if we could associate "labels" with each repo sync type for organizational purposes. We could mark some repo syncs as `git` or `github-api` for instance, or `beta`/`experimental` etc. This would be good to show in the UI (and possibly enable basic filtering on).

This ticket refers to the backend work (schema changes) necessary to support this. We could potentially

- Add a new `mergestat.repo_sync_type_labels` table (for a 1:N relationship)
- Add a new JSON column to the `mergestat.repo_sync_types` to hold an array of labels

patrickdevivo:
Closing this out as implemented by: https://github.com/mergestat/mergestat/pull/665

PR #665, opened 2022-12-19
feat: support labels for repo sync types

riyaz-ali:
Addresses #578 by adding two new tables to manage labels on a sync type.


riyaz-ali:
This PR adds a new field on `repoSyncTypes` type `tagsByTypename`, that is mapping (N:N) between a label and the tagged sync type.

image

patrickdevivo (approved):

PR #672, opened 2022-12-19
feat: updates the `mergestat.user_mgmt_pg_users` view

### LINKED ISSUE #605: Only show users (in `/settings/user-management`) that have a MergeStat role
In some deployments, where there are other existing DB users, we list them in our app (we list all native DB users by default). I propose we instead _only_ list DB users that have a **mergestat** role (one of `mergestat_role_*`) so that we don't clutter the user list.

image

Do we handle this in the frontend with a client side filter? Or in the backend on the view exposing these users?

patrickdevivo:
to exclude DB users that are *not* in one of the pre-set mergestat roles: `'mergestat_role_admin', 'mergestat_role_user', 'mergestat_role_readonly'`

Should be a backend solution that closes #605

### DIFF
--- a/migrations/900000000000016_update_users_view_only_mergestat_roles.up.sql
+++ b/migrations/900000000000016_update_users_view_only_mergestat_roles.up.sql
@@ -30,7 +30,9 @@ CREATE OR REPLACE VIEW mergestat.user_mgmt_pg_users AS (
     -- Maybe there's a better way in the future to filter out the users that are not in one of the mergestat roles
     -- (using a pattern match for instance, to allow for additional roles to be added without this view being updated)
     -- but this works for now.
-    SELECT * FROM users WHERE (memberof && ARRAY['mergestat_role_admin', 'mergestat_role_user', 'mergestat_role_readonly']::name[]) AND users.rolname <> 'mergestat_admin';
+    SELECT * FROM users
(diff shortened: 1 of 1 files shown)


migrations/900000000000016_update_users_view_only_mergestat_roles.up.sql
```diff
@@ -0,0 +1,37 @@
...
+        FROM pg_catalog.pg_roles r
+        WHERE r.rolname !~ '^pg_' AND r.rolcanlogin
+        ORDER BY 1
+    )
+    -- Filter out users that are not in one of the the mergestat roles to reduce noise in the users management table.
+    -- A CTE is used here so that we can access the memberof column in the WHERE clause.
+    -- Maybe there's a better way in the future to filter out the users that are not in one of the mergestat roles
+    -- (using a pattern match for instance, to allow for additional roles to be added without this view being updated)
+    -- but this works for now.
+    SELECT * FROM users WHERE (memberof && ARRAY['mergestat_role_admin', 'mergestat_role_user', 'mergestat_role_readonly']::name[]) users.rolname <> 'mergestat_admin';
```
amenowanna, line 33:
```suggestion
    SELECT * FROM users WHERE (memberof && ARRAY['mergestat_role_admin', 'mergestat_role_user', 'mergestat_role_readonly']::name[]) AND users.rolname <> 'mergestat_admin';
```

amenowanna (changes requested):

amenowanna (approved):

PR #739, opened 2023-01-24
feat(import): Switch From GitHub GraphGQL API to REST API for auto imports

### LINKED ISSUE #540: Consider making the repo auto imports use the v3 GitHub REST API
This would potentially allow us to support GitHub repo auto imports _without_ a token (need to validate this). It would only work for public repos, but it could enable an interesting onboarding path for someone trying out MergeStat without needing to setup a token (and still see data for a lot of repos)

ramiroJCB:
This pr changes the Graphql API approach  for  the REST API  on the auto imports feature . This way  we can import repos without  a GitHub token .There are some sync that won't work without proper validation, those are :

1. PR Commits
2. Pr Reviews
3. Repo Issues
4. Repo metadata
5. Pull Requests
6. Repo Stars

In this content we also reintroduce stdout logging to the importer

To test this  you need to 

1. Log into mergestat
8. go to /repos page and click on  add repositories
9. Choose an user/org with public repos 
10. Those repos should be appearing  after a while

Closes #540


riyaz-ali:
Shouldn't we have patched this in mergestat-lite by updating the `github_org_repos` and `github_user_repos` virtual table 😅 ?

amenowanna:
We know that syncs that use graphql will still be a token so this is expected. @ramiroJCB as you noted this sync does not respect the rate limit. This helps IMO to make a case for making this a generic sync type so we can leverage the queue concurrency. cc @patrickdevivo

.sqlfluff
```diff
@@ -8,7 +8,6 @@ nocolor = False
 dialect = postgres
 # One of [raw|jinja|python|placeholder]
 templater = raw
-# Comma separated list of rules to check, default to all
```
patrickdevivo, line 11: any reason to remove this?

internal/jobs/repo/import.go
```diff
@@ -218,3 +243,111 @@ func difference(existing, new []string) []string {
...
+		}
+		if resp == nil {
+			break
+		}
+
+		if resp.NextPage == 0 {
+			break
+		}
+		opt.Page = resp.NextPage
+		continue
```
patrickdevivo, line 304: I don't think we need this continue since we're just at the end of the loop

internal/jobs/repo/import.go
```diff
@@ -218,3 +243,111 @@ func difference(existing, new []string) []string {
...
+
+		if resp == nil {
+			break
+		}
+
+		if resp.NextPage == 0 {
+			break
+		}
+		opt.Page = resp.NextPage
+		continue
```
patrickdevivo, line 348: same here - unless im missing something?

internal/jobs/repo/import.go
ramiroJCB: we do .The way the api manages their pagination feature is by `resp.NexPage` , we should always loop to the next iteration is the nextPage from the response !=0 otherwise we break the loop in line 300

patrickdevivo (approved):

PR #781, opened 2023-01-27
fix(sqlq): update sqlq to digest 27190246

ramiroJCB:
This should fix the previous version of sqlq 
I will manually close https://github.com/mergestat/sqlq/issues/36


patrickdevivo:
@ramiroJCB I had some trouble running this on a local db that has existing data in it - saw a bunch of error output but wasn't able to get more detail. Can help debug more tomorrow 
image

ramiroJCB:
@patrickdevivo  I tried to replicate this behavior on my local (applying this changes data in the db) and was not able to replicate .
Could you please share the steps you did to make this happen ?

PR #785, opened 2023-01-30
Persist charts

### LINKED ISSUE #771: Persist charts across query executions
As part of our initial testing of the charts we found that we have to re-configure the chart each time we change or re-run the query. What we would like to do is have the charts not have to be configured each time the query is re-run or modified. We will need to create a way to let the user know if the setttings for the chart are no longer valid (e.g. the column is no longe present in the query results).

german-mergestat:
Addressing ephemeral tabs. (Resolves #771)

### DIFF
--- a/ui/src/views/sql-query-editor/index.tsx
+++ b/ui/src/views/sql-query-editor/index.tsx
@@ -1,10 +1,10 @@
 import { useLazyQuery } from '@apollo/client'
 import { Alert, Button, Spinner, Toolbar } from '@mergestat/blocks'
 import { useEffect, useState } from 'react'
+import { ColumnInfo } from 'src/@types'
 import { ExecuteSqlQuery } from 'src/api-logic/graphql/generated/schema'
 import { EXECUTE_SQL } from 'src/api-logic/graphql/queries/sql-queries'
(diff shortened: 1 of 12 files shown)


ui/src/views/sql-query-editor/index.tsx
```diff
@@ -36,6 +37,33 @@ const QueryEditor: React.FC = () => {
...
 
+  const projectionChanged = (columns: ColumnInfo[]) => {
+    const newProjection = columns.map(c => c.name)
+    const previousProjection = projection.map(c => c.name)
+
+    let changed = false
+    newProjection.forEach(p => {
+      if (!previousProjection.includes(p)) {
+        changed = true
+      }
```
patrickdevivo, line 48: you could potentially exit this loop early if you've found a change. So if there's a change in the first item, you don't need to iterate over all the remaining items

german-mergestat: it was improved here [5ad57eb](https://github.com/mergestat/mergestat/pull/785/commits/5ad57eb99c8274658feb61371f637044cfa689f3) 👍

patrickdevivo (approved):

Issue #792, opened 2023-02-03
Add a new syncer, `GIT_REMOTES`


patrickdevivo:
This is a request from a user working with local git repos (files on disk, mounted into the worker) who wants to be able to access the remotes associated with those repos. It's a bit of a niche use case, but worth implementing considering what it unblocks for this person.

Let's add a new sync type, `GIT_REMOTES` which essentially runs `git remote -v` in the context of the repo and uploads the results into the DB. It should not _clone_ the repo first, which is a departure from our current patterns. A clone will "overwrite" the remotes of the original repo. This syncer would therefore only make sense for repos that are directly accessible by the worker (i.e. mounted onto a filesystem the worker can reach).

PR #794, opened 2023-02-03
feat: implement `GIT_REMOTES` repo syncer

patrickdevivo:
Retrieves the remotes of a git repo. Only relevant for local, on-disk repos. Closes [#792](https://github.com/mergestat/mergestat/compare/feat/null/issues/792).

Errors on remote repos:
image

Works for local file path repos:
image

image

### DIFF
--- a/migrations/900000000000024_git_remotes_syncer.up.sql
+++ b/migrations/900000000000024_git_remotes_syncer.up.sql
@@ -6,7 +6,7 @@ CREATE TABLE IF NOT EXISTS git_remotes (
     repo_id uuid PRIMARY KEY,
     name text NOT NULL,
     url text NOT NULL,
-    _mergestat_synced_at timestamptz NOT NULL DEFAULT now()
+    _mergestat_synced_at timestamp with time zone DEFAULT now() NOT NULL
 );
 
 COMMENT ON TABLE git_remotes IS 'table of git repo remotes';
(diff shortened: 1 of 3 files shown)


migrations/900000000000024_git_remotes_syncer.up.sql
```diff
@@ -0,0 +1,18 @@
+BEGIN;
+
+INSERT INTO mergestat.repo_sync_types (type, description, short_name) VALUES ('GIT_REMOTES', 'Retrieves the remotes of a git repo. Only relevant for local, on-disk repos.', 'Git Remotes') ON CONFLICT DO NOTHING;
+
+CREATE TABLE IF NOT EXISTS git_remotes (
+    repo_id uuid PRIMARY KEY,
+    name text NOT NULL,
+    url text NOT NULL,
+    _mergestat_synced_at timestamptz NOT NULL DEFAULT now()
```
amenowanna, line 9: Semantically is the same but we have been using timestamp with time zone

amenowanna (dismissed):

migrations/900000000000024_git_remotes_syncer.up.sql
patrickdevivo: good call - I blame copilot 😄

amenowanna (approved):

PR #748, opened 2023-01-25
feat: implement vendor and provider constructs

riyaz-ali:
This PR adds `vendor` and `provider` constructs as outlined in documents linked in #746. Closes #746

PR #795, opened 2023-02-03
feat: bitbucket initial implementation

riyaz-ali:
This PR adds initial support for Bitbucket. Closes mergestat/internal#129


riyaz-ali:
Builds on #748

.github/workflows/ci-worker.yaml
```diff
@@ -6,10 +6,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Set up Go 1.18
+      - name: Set up Go 1.19
         uses: actions/setup-go@v3
         with:
-          go-version: 1.18.5
+          go-version: 1.19
```
patrickdevivo, line 12: are there other places we should update our declared Go version as well? `go.mod` file or other CI?

riyaz-ali: It is updated in `go.mod` .. not sure if we use any other CI platforms

PR #802, opened 2023-02-03
feat(migrations): Adding saved_queries schema

### LINKED ISSUE #782: Implement backend work for initial saved queries functionality
(Copied from Notion)

**Schema**

- New `saved_queries` table in the `mergestat` schema
    - `id` (UUID)
    - `created_by` (text)
    - `created_at` (timestamp with timezone)
    - `name` (text)
    - `sql` (text)
    - `visualizations` (jsonb) (TBD maybe a better column name)
- We will also need a basic RLS policy:
    - Saved queries where `created_by == current_user()` can be fully managed (edited, deleted, renamed, etc)
    - Saved queries where this is not true for the current user, are read-only (cannot be changed)

******************GraphQL API******************

- The changes above in the schema should be automatically picked up in Graphile API we use, we shouldn’t need additional API work

ramiroJCB:
This change creates the option for user to save queries they think are useful.
This change also brings  new policies within the new table. 

To test the policies please go to _mergestat/scripts/validate-saved-queries-policies.sql_ and follow the steps there.

Closes #782.


migrations/900000000000026_saved_queries.up.sql
```diff
@@ -0,0 +1,27 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS mergestat.saved_queries(
+    id uuid PRIMARY KEY DEFAULT public.gen_random_uuid() NOT NULL,
+    created_by text,
+    created_at timestamp with time zone,
+    name text,
+    sql text,
+    metadata jsonb
+);
```
amenowanna, line 10: we need to add comments for the table and columns

scripts/validate-saved-queries-policies.sql
```diff
@@ -0,0 +1,59 @@
...
+INSERT INTO mergestat.saved_queries (created_by) VALUES('test2');
+
+INSERT INTO mergestat.saved_queries (created_by) VALUES('test3');
+
+    -- Log into your newly created user --
+
+-- Insert row with created_by equal to your current_user name.
+INSERT INTO mergestat.saved_queries (created_by) VALUES('test')
+
+-- You should be able to only select records that are not matching your
```
amenowanna, line 20:
```suggestion
-- You should be able to see al records from all users
```

patrickdevivo (approved):

PR #817, opened 2023-02-09
feat(migrations): sqlq schema support

### LINKED ISSUE #814: Create new tables for mergestat to support sqlq jobs
We need to add  the next tables:
1. queues
2. job_types
3. jobs
4. job_schedule

ramiroJCB:
Here we add support for the future sqlq implementations.
 We needed to add the queues table because graphql being unable to localize  the sqlq schema, @patrickdevivo  idk if you know a way for that to happen, if not we will  keep this queue table in the public schema
 Closes #814


migrations/900000000000029_sqlq_support_schema.up.sql
```diff
@@ -0,0 +1,42 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS mergestat.jobs_types (
```
patrickdevivo, line 3:
```suggestion
CREATE TABLE IF NOT EXISTS mergestat.job_types (
```

suggestion to change the table name

migrations/900000000000029_sqlq_support_schema.up.sql
```diff
@@ -0,0 +1,40 @@
+BEGIN;
+
+CREATE TABLE IF NOT EXISTS mergestat.job_types (
+    id uuid PRIMARY KEY DEFAULT public.gen_random_uuid() NOT NULL,
+    name text NOT NULL,
+    queue text NOT NULL,
+    description text,
+    default_parameters jsonb,
+    priority int,
+    schedule text,
```
ramiroJCB, line 10: with @patrickdevivo we discuss that we would like this to be a cron string instead of a int or timestamp , that way we would have more freedom to manipulate the way we handles schedules

patrickdevivo (approved):

Issue #843, opened 2023-02-16
Add description column to 'Providers'


amenowanna:
We would like to give our users the ability to add more context to a Git Source by adding a description column to the providers table. This is something we allow the users to manage with configuring a Git Source. One of the example use cases could be to distinguish between Git Sources of the same vendor type. For example, `This Git Source uses the PAT for Org A which needs to be segmented for reasons X and Y`

Issue #828, opened 2023-02-13
Refactor new `vendor`/`provider` schema


patrickdevivo:
A couple of items from our discussion on the `vendor`/`provider` work that's been happening.

- [x] Add a `type` column in the `vendors` table to support when we extend beyond "Git Sources" as the provider type. First type is `GIT_SOURCE` (or similar)
- [x] Move the `username` column on `service_auth_credentials` table into the `credentials` column (to be a part of the JSONB). This would allow providers that don't have a username to be a bit less clumsy - or providers that have another field like `clientID` or something like that

PR #847, opened 2023-02-17
fix: update vendor-provider schema

riyaz-ali:
This PR addresses #828 and #843


migrations/900000000000042_add_comment_to_providers.up.sql
```diff
@@ -0,0 +1,6 @@
+-- SQL migration to add an optional comment field to providers
+BEGIN;
+
+ALTER TABLE mergestat.providers ADD COLUMN comment TEXT;
```
patrickdevivo, line 4:
```suggestion
-- SQL migration to add an optional description field to providers
BEGIN;

ALTER TABLE mergestat.providers ADD COLUMN description TEXT;
```

patrickdevivo: @riyaz-ali any reason for the comment vs description?

riyaz-ali: no particular reason 😅 happy to rename it to `description` 👍

migrations/900000000000043_encrypt_username_in_crednetials.up.sql
```diff
@@ -0,0 +1,35 @@
+-- SQL migration to update credential add and fetch routines to perform encryption and decryption
```
patrickdevivo, line 1: just a nit - but see typo in file name `crednetials` @riyaz-ali

PR #860, opened 2023-02-21
fix: only highlight one item in the navbar at a time

### LINKED ISSUE #840: Only highlight the child item in the navbar
When a "child" navigation item is selected, only highlight that (not the parent as well). This should make it clearer what page the user is on.

image

patrickdevivo:
Closes #840

image


german-mergestat (approved):
LGTM

PR #862, opened 2023-02-23
fix(editor): checking for rows affected in query editor

### LINKED ISSUE #824: Odd/incorrect behavior when query returns 0 rows of data
Seeing some odd behavior when a SQL execution returns no rows of data. For some reason I'm consistently seeing `63 rows impacted` when there are no rows returned. I think we may need to take a look at this logic again/

image

ramiroJCB:
So we werent checking for rows counts / rows affected  so when we did and operation with no rows returned we ommited the result and  show the empty editor when actually we had rows affected.
With this changes we take in consideration all rows and rows count/affected.

To test you need to :
1. do any operation that returns no rows and also affect rows
2. you should see the rows impacted but that query

screen_shot_2023-02-23_at_11 21 22_am_720

Closes #824

### DIFF
--- a/ui/src/views/sql-query-editor/components/state-rows-affected.tsx
+++ b/ui/src/views/sql-query-editor/components/state-rows-affected.tsx
@@ -1,18 +1,18 @@
 import { Avatar } from '@mergestat/blocks'
 import { MinusIcon } from '@mergestat/icons'
-import { QueryResultProps } from 'src/@types'
+import { ExecuteSqlQuery } from 'src/api-logic/graphql/generated/schema'
 
 type QueryEditorRowsImpactedProps = {
-    data: QueryResultProps | undefined
-    children?: React.ReactNode
+  data: ExecuteSqlQuery | undefined
+  children?: React.ReactNode
 }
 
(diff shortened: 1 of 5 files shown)


ui/src/views/sql-query-editor/components/state-rows-affected.tsx
```diff
@@ -0,0 +1,18 @@
...
+type QueryEditorRowsImpactedProps = {
+  data: ExecuteSqlQuery | undefined
+  children?: React.ReactNode
+}
+
+const QueryEditorRowsImpacted: React.FC<QueryEditorRowsImpactedProps> = ({ data }: QueryEditorRowsImpactedProps) => (
+  <div className='flex-1 h-1/2 flex flex-col items-center justify-center p-8'>
+    <Avatar icon={<MinusIcon className="t-icon" />} />
+    <p className='t-text-muted mt-6'>
+      {`rows impacted ${data?.execSQL.rowCount}`}
```
patrickdevivo, line 14:
```suggestion
      `${data.execSQL.rowCount} row${data.execSQL.rowCount > 1 ? 's' : ''} impacted`
```

patrickdevivo (changes requested):

PR #927, opened 2023-03-11
chore(gitlab): Adding GitLab flow

german-mergestat:
Adding first approach for `GitLab` flow.

Screenshot 2023-03-11 at 3 56 59 PM

### DIFF
--- a/ui/src/utils/index.tsx
+++ b/ui/src/utils/index.tsx
@@ -272,3 +272,12 @@ export const getVendor = (vendor: string) => {
 export const getVendorLabelType = (vendor: string) => {
   return (vendor === VENDOR_TYPE.GITHUB || vendor === VENDOR_TYPE.BITBUCKET) ? 'organization' : 'group'
 }
+
+/**
+ * Method to check if repo start with http
+ * @param repo Repo url to validate
+ * @returns same url repo if it stats with http, otherwise returns undefined
+ */
+export function getExternalRepoLink(repo?: string) {
+  return repo && /^http/ig.test(repo) ? repo : undefined
+}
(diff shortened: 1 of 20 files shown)


ui/src/views/repo-git-sources/add-source/index.tsx
```diff
@@ -126,13 +126,13 @@ const AddSourceView: React.FC = () => {
                   onChange={() => setVendor(VENDOR_TYPE.BITBUCKET)}
                   startIcon={<BitbucketIcon className="t-icon" />}
                 />
-                {/** <RadioCard
+                {<RadioCard
                   isSelected={vendor === VENDOR_TYPE.GITLAB}
                   label="Gitlab"
```
patrickdevivo, line 131:
```suggestion
                  label="GitLab"
```

ui/src/utils/index.tsx
```diff
@@ -40,6 +40,15 @@ export function validateGitHubPAT(token: string) {
   return /^ghp_[a-zA-Z0-9]{36}$/ig.test(token)
 }
 
+/**
+ * Method to validate given token is a right GitLab PAT
+ * @param token Token string to validate
+ * @returns true if token is a right GitLab PAT, otherwise return false
+ */
+export function validateGitLabPAT(token: string) {
+  return /^glpat-[a-zA-Z0-9]{36}$/ig.test(token)
```
patrickdevivo, line 49:
```suggestion
  return /^glpat-[a-zA-Z0-9]{20}$/ig.test(token)
```

I'm not sure if this is true of all tokens, but the one I created for testing was 20 characters after the `glpat-` prefix. If we don't find anything official in the docs about the length/size of the tokens, I think the only validation we should do is checking for the prefix - and maybe eventually an API call

patrickdevivo:
When adding a GitLab org auto import, the name of the org doesn't appear in the list:

image

and when the auto import tries to run:
image

So I think we might need to fix the import type here

patrickdevivo (changes requested):

patrickdevivo:
image

patrickdevivo (approved):

PR #929, opened 2023-03-13
fix: address naming conflict in `graphile` with schema update

patrickdevivo:
Should fix the error appearing here:

image

@riyaz-ali let me know your thoughts on name, not tied to this.


riyaz-ali:
Well, we no longer need `mergestat.jobs` and `mergestat.job_types` tables. I'd let @amenowanna and @ramiroJCB confirm this. If yes, then it'd be better to drop them rather than adding this comment.

amenowanna:
They can go based on what we discussed

PR #934, opened 2023-03-13
feat(graphql): only counting syncs that have scheduleEnabled true

### LINKED ISSUE #913: Sum of repo syncs doesn't always add up correctly
Sometimes we show an incorrect sum in the `/repos` table:

Image

ramiroJCB:
Closes #913

### DIFF
--- a/ui/src/api-logic/graphql/queries/get-repos.ts
+++ b/ui/src/api-logic/graphql/queries/get-repos.ts
@@ -65,7 +65,7 @@ const GET_REPOS = gql`
           vendor
           settings
         }
-        repoSyncs(orderBy: SYNC_TYPE_ASC, condition:{scheduleEnabled:true}) {
+        repoSyncs(orderBy: SYNC_TYPE_ASC,filter:{lastCompletedRepoSyncQueueId:{isNull:false}}) {
           totalCount
           nodes {
             id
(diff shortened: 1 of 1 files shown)


ui/src/api-logic/graphql/queries/get-repos.ts
```diff
@@ -65,7 +65,7 @@ const GET_REPOS = gql`
           vendor
           settings
         }
-        repoSyncs(orderBy: SYNC_TYPE_ASC) {
+        repoSyncs(orderBy: SYNC_TYPE_ASC, condition:{scheduleEnabled:true}) {
```
german-mergestat, line 68:
```suggestion
        repoSyncs(
          orderBy: SYNC_TYPE_ASC
          filter: {lastCompletedRepoSyncQueueId: {isNull: false}}
        ) {
```

german-mergestat: This way it works... and run `npm run codegen` to update generated schema on the UI (_/ui/src/api-logic/graphql/generated/schema.ts_).

german-mergestat (changes requested):

ui/src/api-logic/graphql/queries/get-repos.ts
ramiroJCB: done

PR #938, opened 2023-03-14
fix: `grype` and `gitleaks` scanners had some perms/misconfig issues

patrickdevivo:
validated locally. see diff for summary of fixes. my guess is recent `Dockerfile` changes caused the issue with Gitleaks at least.

### DIFF
--- a/internal/syncer/grype_repo_scan.go
+++ b/internal/syncer/grype_repo_scan.go
@@ -26,7 +26,7 @@ func (w *worker) handleGrypeRepoScan(ctx context.Context, j *db.DequeueSyncJobRo
 		}
 	}()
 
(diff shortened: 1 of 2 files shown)


internal/syncer/grype_repo_scan.go
```diff
@@ -26,7 +26,7 @@ func (w *worker) handleGrypeRepoScan(ctx context.Context, j *db.DequeueSyncJobRo
 		}
 	}()
 
-	if err = w.clone(ctx, j.Repo, j); err != nil {
```
ramiroJCB, line 29: Interesting, How this hapenned

ramiroJCB (approved):

PR #989, opened 2023-04-04
fix(filters) Solving bad behavior in tables filter and keeping filter…

### LINKED ISSUE #984: When paging /repos we do not persist the number of records or page we were on
When paging through the results of /repos the number of records per page is not persisted when we drill into a repo and then go back to the /repos experience. Likewise, when we have paged through the results when we go back we are taken back to the first pages of results and not the page we were on previously.

Image

german-mergestat:
Solving bad behavior in `tables filter` and keeping filters for `repo` page. (Resolves #984)


amenowanna (approved):
Love this fix! ❤️

PR #1021, opened 2023-04-18
Adding auto imports for container sync

### LINKED ISSUE #1019: Add container sync support for default syncs in auto imports
We need to create a /repos-containers/git-source/{uuid} page where we can implement default syncs for container sync auto imports. 

Relates to #1008

amenowanna:
Resolves #1019 

- [x] Added udf for adding a new repo import for container syncs
- [x] Added udf to update default container images for repo 
- [x] Created new git-sources pages for container syncs that use the new udf's to manage the repo imports
  - It is accessed through `/repos-containers/git-sources/<id-gitsource>`
- [x] Updated worker to handle new `defaultContainerImages` in repo imports
- [x] Fix latest sync status bars and related status issues
- [x] Add container images seed values to DB
- [x] Fix priority issues with GitHub syncs and Repo Imports

### DIFF
--- a/migrations/900000000000055_add_repos_stats_functions.up.sql
+++ b/migrations/900000000000055_add_repos_stats_functions.up.sql
@@ -171,6 +171,7 @@ BEGIN
             ci.name AS container_image_name,
             j.id AS job_id,
             j.status,
+            j.created_at,
             j.started_at,
             j.completed_at,
             (SELECT COUNT(1) FROM sqlq.job_logs WHERE sqlq.job_logs.job = j.id AND level = 'warn') warning_count,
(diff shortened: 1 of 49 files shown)


migrations/900000000000055_add_repos_stats_functions.up.sql
amenowanna, line 1: @german-mergestat can you drop this change. We will need to add it to the latest migration since this migration is already in the wild.

migrations/900000000000058_add_default_syncs_for_container_images.up.sql
```diff
@@ -186,4 +186,40 @@ BEGIN
...
+            j.started_at,
+            j.completed_at,
+            (SELECT COUNT(1) FROM sqlq.job_logs WHERE sqlq.job_logs.job = j.id AND level = 'warn') warning_count,
+            (SELECT COUNT(1) FROM sqlq.job_logs WHERE sqlq.job_logs.job = j.id AND level = 'error') error_count
+        FROM mergestat.container_syncs cs
+        INNER JOIN mergestat.container_sync_executions cse ON cs.id = cse.sync_id
+        INNER JOIN mergestat.container_images ci ON cs.image_id = ci.id
+        INNER JOIN sqlq.jobs j ON cse.job_id = j.id
+        WHERE cs.id = container_syncs.id
+        ORDER BY cs.id, j.completed_at DESC
```
german-mergestat, line 212: @amenowanna Should this commit have the `created_at` here in't it? and I am thinking it is enough just sorting from here by `created_at` and in the fronted it wouldn't be necessary

german-mergestat: I just remembered `JSONB_OBJECT_AGG` changes the order jeje, so it is still necessary apply the sorting from Frontend.

amenowanna: Good catch on both! I i didn't notice any issues with the sort though. I just pushed my fix.

migrations/900000000000058_add_default_syncs_for_container_images.up.sql
```diff
@@ -0,0 +1,205 @@
...
```
german-mergestat, line 192:
```suggestion
        ORDER BY cs.id, j.created_at DESC
```
@amenowanna it is important order by `created_at` here because it takes first 15 records, so we need to sort by `created_at` here as well to get it working

amenowanna: Good call!

Issue #1020, opened 2023-04-18
github_repo_info not populated


mattheworiordan:
Firstly thanks for an awesome project 👋 

I've recently stated using mergestat and I've found that many of my repose are not appearing in Grafana. When I looked into the Grafana query, I can see that it does a join with github_repo_info:

```
SELECT owner || '/' || name FROM github_repo_info JOIN repos ON github_repo_info.repo_id = repos.id
```

What is confusing is that the table `github_repo_info` is only partially populated with what seems like potentially just public repos, although not exclusively (other public repos are not in that table).

Do you have any idea what may have caused that table to not be populated?  `repos` contains a list of all repositories...

Is there any way to force a refresh/reload?

Thanks for your help.

amenowanna:
@mattheworiordan the table is populated by a GitHub  Metadata/Info sync. When we first created those dashboards GitHub was central to our solution. We need to refactor those dashboards to not have that dependency so they will work even when that data is not present. We have been thinking of removing the built in grafana so have not prioritized this. If you are finding this to be valuable we would be interested to know to evaluate the longevity of grafana in the project.

mattheworiordan:
> the table is populated by a GitHub Metadata/Info sync

Ok, good to know. Is there some way I can retry the Github Metadata/Info sync OOI?  It's odd that it's partially complete, but if I could manually trigger a job I could try and inspect any logs to understand what is going on.

> We have been thinking of removing the built in grafana so have not prioritized this. If you are finding this to be valuable we would be interested to know to evaluate the longevity of grafana in the project.

Right now it seems of interest, but quite honestly it's too early days. I will perhaps make some changes to remove the dependency on `github_repo_info` so that they work, and if useful can issue a PR to bring the changes back?

amenowanna:
@mattheworiordan The plan for now is to remove grafana from the repo in favor of running and viewing queries inside the MergeStat UI. That being said we would welcome the feedback on why this may not be a good idea. To enable this sync across all your repos we currently do not have a bulk update experience but one is about to come up with a new syncing approach we are taking. But here is a query you can run to enable the sync across all the repos. Several assumptions here include that these are GitHub repos and you have added a Personal Access Token for GitHub. If that is not the case then you will need to update the queries to remove this table from the dependencies. PR's are always welcome and if you find value in the grafana boards please let us know.

```sql
INSERT INTO mergestat.repo_syncs (repo_id, sync_type, priority, schedule_enabled)
SELECT
[...]

PR #1026, opened 2023-04-20
chore: comment out `grafana` for now

patrickdevivo:
since we haven't been investing in it, and our approach has changed a bit since. Hopefully should avoid some confusion.


amenowanna:
All the examples are still in the repo and can be used in a Grafana instance so I am good with this. Also mentioned this pr in #1020

amenowanna (approved):

PR #1048, opened 2023-05-02
Demo role and statement timeouts

amenowanna:
- [x]  Create a new `mergestat_demo` role
    - [x]  Update `mergestat.user_mgmt_update_user_password` UDF to error when `mergestat_demo` role is calling it
    - [x]  Update `execSQL` custom mutation to enforce readonly tx when `mergestat_demo` is the current role. Check if current user is member of the `mergestat_demo` role.
- [x]  Lockdown queries with tighter timeouts
    - [x]  `execSQL` using `set_statement_timeout` in the tx

### DIFF
--- a/graphql/exec-sql/index.ts
+++ b/graphql/exec-sql/index.ts
@@ -76,6 +76,9 @@ module.exports = makeExtendSchemaPlugin({
             );
         }
 
+        //Then set the session timeout
+        await context.pgClient.query("SET statement_timeout = '30s';")
+
         // then create a cursor https://node-postgres.com/api/cursor for the user supplied query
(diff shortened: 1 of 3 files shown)


graphql/exec-sql/index.ts
```diff
@@ -50,6 +50,34 @@ module.exports = makeExtendSchemaPlugin({
...
+                )
+                THEN SET TRANSACTION READ ONLY;
+                END IF;
+            END
+            $do$;`
+            );
+        }
+
+        //Then set the session timeout
+        await context.pgClient.query("SET statement_timeout = '30s';")
```
patrickdevivo, line 80: eventually maybe we allow this to be set via env

patrickdevivo (approved):

PR #1063, opened 2023-05-06
chore: removes unused code (v1) from the worker

patrickdevivo:
Should also improve build times, as `libgit2` and `sqlite` are no longer dependencies.


amenowanna:
@riyaz-ali FYI...

riyaz-ali:
looks good! we can go ahead with this if we want to deprecate older syncs and move ahead with a container-only approach 👍

amenowanna:
@riyaz-ali The main thing I see missing with this PR is a migration to drop the table that are no longer needed. Can you please add this.

riyaz-ali:
Let's add that migration in another PR? I want to ensure that the changes made here in this PR are working before making any destructive / data-loss operation.

PR #1076, opened 2023-05-16
fix(schema) Solving error even when routing to another age diff…

german-mergestat:
Solving error even when routing to another age different to `/login`.


amenowanna (approved):
Works great! Doesn't call the schema introspection until after login too.

ui/src/hooks/useMonacoProvider.tsx
```diff
@@ -9,7 +9,7 @@ import { sqlKeywords } from 'src/utils/monaco-editor/autocomplete/sql-keywords'
 export const useMonacoProvider = () => {
```
Morton22, line 9: ui/src/hooks/useMonacoProvider.tsx

PR #1079, opened 2023-05-19
Saved explores and Explore UDF Updates

amenowanna:
- Update explore udf to fix new parameters that we swapped. Mark function as STABLE so it can be used as a query instead of a mutation.
- Add saved explores schema and manual tests
- Fixed uppercase for role in saved queries manual tests
- Updated sqlq


german-mergestat:
Addressing following points: 

- Change explore_ui to a `STABLE` udf which will require the UI to consume as a query and not a mutation. This should allow postgres to create better plans for the queries too.
- Committer when and/or author when filters.
- Query results are returning 1001 and not 1000 results (In `Explore Modal` as well as `Query Editor` page).
- Modals are not giving a `warning` if there are more than 1000 results like we have in the queries experience.

alert-rows-exceded

patrickdevivo (dismissed):

german-mergestat:
Addressing following points:

- Don’t show any charts just some sort of empty state like we do in /repos when we don’t have any repos.
- Make Explore a top level Nav element and make it the primary one.
- Number of repos indexed by explore piece of UI/query.
- Manage links directly to the “special” explore sync, which we’ll need a way of fetching the UUID for.
- Optimizing a little the performance using `useMemo()` hook in some parts of the application.

explore-page

PR #1082, opened 2023-05-23
feat: add encrypted store for user-defined variables for sync

riyaz-ali:
This PR adds an encrypted storage table to store user-provided variables for container-based sync.


migrations/900000000000064_add_variable_store_for_container_sync_jobs.up.sql
```diff
@@ -0,0 +1,39 @@
...
+    PRIMARY KEY (repo_id, key)
+);
+
+
+-- encrypt and store the given variable
+CREATE OR REPLACE FUNCTION mergestat.add_sync_variable(repo_id UUID, key TEXT, value TEXT, secret TEXT)
+RETURNS mergestat.SYNC_VARIABLES AS $$
+DECLARE _inserted_row mergestat.sync_variables;
+BEGIN
+    INSERT INTO mergestat.sync_variables(repo_id, key, value)
```
amenowanna, line 22: This should probably be an upsert to handle conflicts

riyaz-ali: Ideally, I'd want the user to explicitly delete the variable and re-create. This way (in future) if they've accidentally exposed it to a sync they didn't want to, deleting the variable would purge that. Although, the user would need to go ahead and add it again to all the other syncs that they've granted access to.

amenowanna (approved):

PR #1122, opened 2023-10-08
feat: add possibility to set scheduler and syncer interval

griseau:
Asked on the [slack](https://mergestatcommunity.slack.com/archives/C02KEJ5M7RA/p1696672369445719) how to configure how often jobs are scheduled, so started looking at it. 

It seems that it was already mentioned in the code, so just adding env variable with default values to make it possible to run only once a day for example (In my case we have a lot of repos, running it more often doesn't really bring value but add a lot of computing). 

I didn't configure the timeout one, but can add it if you think it's necessary


griseau:
Long time I didn't write go 😅 
Should be all good now !

patrickdevivo (approved):

PR #1123, opened 2023-10-10
fix: set container sync on the same schedule as scheduler

griseau:
Previous PR #1122 wasn't working as I missed the cron, that I guess needs to be set as the same value as the scheduler ?

### DIFF
--- a/cmd/worker/main.go
+++ b/cmd/worker/main.go
@@ -295,7 +295,7 @@ func main() {
 	go cron.AutoImport(ctx, 15*time.Second, upstream)
 
 	// run container sync scheduler every minute
-	go cron.ContainerSync(ctx, 1*time.Minute, upstream)
+	go cron.ContainerSync(ctx, time.Duration(schedulerInterval)*time.Minute, upstream)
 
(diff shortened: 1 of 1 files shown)


cmd/worker/main.go
```diff
@@ -295,7 +295,7 @@ func main() {
 	go cron.AutoImport(ctx, 15*time.Second, upstream)
 
 	// run container sync scheduler every minute
-	go cron.ContainerSync(ctx, 1*time.Minute, upstream)
+	go cron.ContainerSync(ctx, time.Duration(schedulerInterval)*time.Minute, upstream)
```
patrickdevivo, line 298: I believe this should be the syncer interval
```suggestion
	go cron.ContainerSync(ctx, time.Duration(syncerInterval)*time.Minute, upstream)
```

patrickdevivo (changes requested):

cmd/worker/main.go
griseau: This process is the one updating the queue of jobs right ? 
schedulerInterval is in minute whereas syncerInterval is in seconds

PR #1160, opened 2024-03-19
Use docker api instead of podman cli. Fixes #1075

### LINKED ISSUE #1075: Investigate (and implement) if it's possible to run the `worker` in non-privileged mode
For some users, running the `worker` container in privileged mode is a no-go. We should investigate if and how we can allow for non-privileged mode use with our container-sync approach (podman in a container)

cvhariharan:
This change allows running mergestat without privileged mode.


sleipnir:
Hey guys any change to see this merged on main? Great job @cvhariharan

PR #1198, opened 2025-10-06
Fix container sync execution in Docker environments

santosfh:
## Problem

Container-based syncs fail to execute when the worker runs inside Docker with Podman, resulting in:
```
exec container process '/syncer/entrypoint.sh': Invalid argument
```

This occurs because rootless Podman cannot properly execute containers when running inside a Docker container due to user namespace and cgroup delegation issues.

## Solution

This PR switches from Podman to Docker CLI for container execution, which works correctly in nested containerization scenarios.

### Changes:
- **Dockerfile**: Replace `podman fuse-overlayfs` with `docker-cli`
- **docker-compose.yaml**: 
  - Mount Docker socket from host
  - Add shared git repository volume with matching paths
  - Run worker as root for Docker socket access
  - Set `GIT_CLONE_PATH` environment variable
- **podman.go**:
  - Auto-detect and prefer Docker over Podman
  - Remove `docker://` prefix from image URLs
  - Remove `--restart` flag (conflicts with `--rm` in Docker)
  - Change network from `host` to `mergestat_default`
  - Support both Docker and Podman image metadata formats

## Testing

Tested on Docker Desktop with successful sync execution. Syncs now complete successfully with proper git cloning and database operations.

## Compatibility

The changes maintain backward compatibility with Podman - if Docker is not available, the system will fall back to Podman.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/usr/share/pkgconfig/libgit2/lib/pkgconfig/ mak
 FROM zricethezav/gitleaks:v8.15.3 AS gitleaks
 
 FROM alpine:3.18
-RUN apk upgrade && apk add --no-cache curl postgresql-client ca-certificates git go podman fuse-overlayfs tini openssl1.1-compat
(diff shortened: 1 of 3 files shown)


santosfh:
Closing - this PR should go to the fork instead of the original stale repository.

